PR #452, opened 2016-02-03
451 scenarioo validator

xeronimus:
@bruderol First, I had the validator within the scenarioo-server project. but since I'd like to use it as a standalone tool (gradle "application" plugin) it was easier to implement this a s a own gradle project.

Please also note the two methods I made public in order to reuse code.
* org.scenarioo.business.builds.ScenarioDocuBuildsManager#loadBuildImportSummaries
* org.scenarioo.business.builds.ScenarioDocuBuildsManager#loadBranchBuildsList

what do you think?

### DIFF
--- a/scenarioo-validator/src/main/java/org/scenarioo/validator/ScenariooValidator.java
+++ b/scenarioo-validator/src/main/java/org/scenarioo/validator/ScenariooValidator.java
@@ -1,4 +1,4 @@
-package org.scenarioo.tools.validator;
+package org.scenarioo.validator;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.filefilter.RegexFileFilter;
@@ -14,9 +14,7 @@ import java.util.Map;
 
 /**
  * Validates a given docu directory by trying to import it.
- * If import (parsing + aggregation) was successful, the produced folder structure and xml files within the given directory are valid.
- * <p/>
- * This can also be used as command line tool in order to validate any scenarioo documentation folder (e.g. produced by another writer API like scenarioo-js)
+ * If import (parsing + aggregation) was successful, the folder structure and xml files within the given directory are valid.
  */
 public class ScenariooValidator {
 
@@ -46,6 +44,7 @@ public class ScenariooValidator {
             return;
         }
 
+        // Note: it would be sufficient to delete "version.derived.properties". But we like to start with a clean slate.
         Collection<File> derivedFiles = FileUtils.listFiles(this.docuDirectory, new RegexFileFilter("(.*derived\\.(xml|log))|(.*derived\\.properties)"), TrueFileFilter.INSTANCE);
         LOGGER.info("Cleaning " + derivedFiles.size() + " derived files from docu directory " + this.docuDirectory);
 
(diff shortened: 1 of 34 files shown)


scenarioo-validator/src/main/java/org/scenarioo/validator/ScenariooValidator.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public boolean validate() throws InterruptedException {
+        cleanDerivedFilesIfRequested();
+        Map<BuildIdentifier, BuildImportSummary> buildImportSummaries = validationBuildImporter.importBuildsAndWaitForExecutor();
+        return evaluateSummaries(buildImportSummaries);
+    }
+
+    /**
+     * will remove all derived files in given docuDirectory (if requested).
+     * removes "*.derived.xml"  and "*.derived.log"  files.
```
bruderol: also mention *.derived.properties" file!
Or why not just "*.derived.*"?

scenarioo-validator/src/test/java/org/scenarioo/validator/ScenariooValidatorTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.io.File;
+
+public class ScenariooValidatorTest {
+
+    /**
+     * Our validator must successfully validate our own example docu.
+     */
+    @Test
+    @Ignore("This is ignored since this test produces derived files in scenario-docu-generatio-example which is under version control")
```
bruderol: hmm, yes, we should start to implement better tests in server project and for validator as well.

Idea: we could introduce some test samples that we can import using the importer and the validator (using the clean function we could always delete the derived files and of course we could not refer to the files that are checked in but use the resources in the build folder instead, where it does not matter if something is changed). Like this we could easily improve tests in validator and in server (also for testing backward compatibility issues).

I open an issue for this with a proposal: #453

bruderol:
Just great! :+1: Very helpful contribution. :smile: 
I am also happy to see, that it was quite easy to implement this from what we allready had.
Thank you so much!

Just two things:
1. How will we release the tool? Will it be included in the WAR (even though it is not part of the server project)?
2. Can you please document somewhere (probably in Wiki) that this tool exists and how to use (=start) it?

From my side, you can merge, but I want leave it open to you whether you want fix the minor things mentioned above before or after merging.

@adiherzog Any comment from your side?

bruderol:
@xeronimus just one minor feedback: If you did the feature branch directly inside our scenarioo repos, I could have fixed the missing comment for "*,derived.properties" directly on the branch in the PR-review pane. So it would help in future if you open your feature branches directly inside the scenarioo repos to better collaborate on features.

xeronimus:
Hi. thanks for your feedback. I just fixed some minor issues and will then document and merge.

xeronimus:
@bruderol  regarding your question "How will we release the tool". Quick answer:
the gradle build produces a runnable script (for UNIX and windows)

`$ ./gradlew clean build installApp`
`$  ./scenarioo-validator/build/install/scenarioo-validator/bin/scenarioo-validator -c /path/to/my/docuDir`

long answer will be in the docu.

bruderol:
ok, I think we should provide a binary package that one can use without the need of gradle (should be very easy, since we allready have continuous integration builds). But ok, for now this is sufficient.

Issue #446, opened 2015-12-16
HTML View is broken in 2.1.0


danielsuter:
HTML code won't be shown in version 2.1.0. This was most probably broken with the screen annotation release. The step.xml files contain the html code but it won't be displayed in the scenarioo viewer web application.

You can reproduce this issue in the scenarioo demo version.
1. http://54.88.202.24/scenarioo-master/#/step/Donate/find_donate_page/searchResults.jsp/0/0?branch=wikipedia-docu-example&build=last%20successful
2. Open HTML tab

bruderol:
I put this to academy and ready to code, because urgent. Would be great if somebody could fix this together with #454

bruderol:
@mi-we I will remove this branch. Please take care to remove remote branches when merged to develop. Thanks

Issue #430, opened 2015-09-29
Do not show the HTML tab on a step, if the HTML field is empty


adiherzog:

bruderol:
@mi-we Can we close this now?

mi-we:
@bruderol Yep, I will close the issue.

PR #455, opened 2016-03-01
#446 HTML View is broken in 2.1.0

mi-we:
fixed mapping of htmlSource from step data to frontend


mi-we:
The fix itself was rather quick, I will also try to modify one of the e2e tests to check, that html returned by the service is displayed

mi-we:
I saw that #430 was also concerning the HTML tab so in accordance with @adiherzog  I also pushed that solution to this PR.

bruderol:
:+1: Thanks! :smile:

Issue #447, opened 2015-12-21
Page numbers are wrong in version 2.1


adiherzog:
Page number is always 1, which is wrong. Instead, pages should be numbered 1, 2, 3, etc.

![page numbers wrong](https://cloud.githubusercontent.com/assets/3780183/11941692/bde1f518-a831-11e5-8126-e46f7853b957.PNG)

bruderol:
@mi-we Thanks for the fix! Great solution!

PR #463, opened 2016-03-15
#447 Fixed numbering of pages on scenario overview page

mi-we:


bruderol:
Hi @mi-we and @adiherzog 
I assigned this to me, because I had shortly time to review it and the bug was probably even a mistake by me. Thank you for the very clean solution. :+1: :100: TOP!
I merge it.

PR #481, opened 2016-04-12
Merge bug fix from release-2.2

bruderol:


bruderol:
@adiherzog Maybe you can review before we also merge it to develop?? Thank you.

PR #493, opened 2016-04-19
Fixed initial table sort by scenario.name in scenario table in usecase.js

patrickschweizer:
…Case.js


bruderol:
@adiherzog @patrickschweizer 
If I understand this right, this is a fix, that we should merge into the release 2.2 - @adiherzog do you agree, if we merge that small fix?

bruderol:
@adiherzog I will merge that now. Think it is okay.

adiherzog:
Thanks for the fix and for merging. Looks good :+1:

bruderol:
Tataaaaa .... ready for the release :+1:

Issue #509, opened 2016-05-09
John Papa: Introduce VM pattern in remaining controllers


adiherzog:
We startend introducing the vm (view model) pattern of John Papa last year. Now we should finish this for all remaining controllers.

PR #540, opened 2016-07-14
Full text search

adiherzog:
@bruderol Please review and let me know when I can merge it. If it's basically ok we can merge it and still do improvements later (like e.g. performance improvements).


scenarioo-client/app/manage/generalSettings/generalSettings.controller.js
```diff
@@ -17,13 +17,14 @@
...
 
-function GeneralSettingsController($scope, BranchesResource, ConfigService) {
+function GeneralSettingsController($scope, BranchesResource, ConfigService, SearchEngineStatusService) {
 
     var vm = this;
     vm.branches = [];
     vm.configuration = {};
     vm.configuredBranch = {};
     vm.successfullyUpdatedConfiguration = false;
+    vm.isEngineRunning = false;
```
bruderol: Engine - what kind of engine ?? please rename to something like "isSearchEngineRunning"

scenarioo-client/app/manage/generalSettings/generalSettings.controller.js
```diff
@@ -55,6 +57,12 @@ function GeneralSettingsController($scope, BranchesResource, ConfigService) {
         }
     }
 
+    function getEngineRunning () {
```
bruderol: is no getter, does not even return what is loaded here - should be called something like `loadSearchEngineRunning`.

scenarioo-client/app/shared/navigation/navigation.controller.js
```diff
@@ -31,6 +31,22 @@ function NavigationController($scope, $location, LocalStorageService, BranchesAn
 
     SelectedBranchAndBuildService.callOnSelectionChange(loadBranchesAndBuildsService);
 
+    $scope.globalSearch = {
```
bruderol: why do we still use $scope here - shouldnt we use `vm` and `controllerAs` notation now?

scenarioo-client/app/shared/navigation/navigation.controller.js
```diff
@@ -31,6 +31,22 @@ function NavigationController($scope, $location, LocalStorageService, BranchesAn
...
+        queryString: ''
+    };
+
+    $scope.search = function () {
+        $location.path('/search').search('q=' + $scope.globalSearch.queryString);
+    };
+
+    function getEngineRunning () {
```
bruderol: see other comment concerning wrong getter naming and "Engine" naming.

scenarioo-client/app/shared/searchEngineStatus.service.js
```diff
@@ -0,0 +1,38 @@
...
+
+angular.module('scenarioo.services').factory('SearchEngineStatusService', function (SelectedBranchAndBuildService, FullTextSearchService) {
+    return {
+        isEngineRunning: function() {
+            var selected = SelectedBranchAndBuildService.selected();
+
+            var promise = FullTextSearchService.checkStatus({
+                    buildName: selected.build,
+                    branchName: selected.branch
+                }
```
bruderol: not understand why this depends on current branch and build selection?? will it report false if index is missing?

scenarioo-client/app/manage/generalSettings/generalSettings.html
```diff
@@ -54,6 +54,17 @@
             </div>
         </div>
 
+        <div class="row sc-space-bottom">
+            <div class="col-lg-3">Search engine
+                &nbsp;<span class="glyphicon glyphicon-info-sign" tooltip="Application specific full text search engine."></span>
+            </div>
+            <div class="col-lg-9" ng-show="!vm.isEngineRunning" ng-cloak>
+                <div class="alert alert-warning">No search engine running</div>
```
bruderol: Please add a link to our wiki here, that explains how to install / configure it.

scenarioo-client/app/manage/generalSettings/generalSettings.html
```diff
@@ -54,6 +54,17 @@
...
+            </div>
+            <div class="col-lg-9" ng-show="vm.isEngineRunning" ng-cloak>
+                <span>up and running!</span>
```
bruderol: on the demo I see this text, but no search field for searching - is that as expected?

scenarioo-server/build.gradle
```diff
@@ -25,11 +25,11 @@ dependencies {
     compile 'org.jboss.resteasy:resteasy-jaxb-provider:2.3.3.Final'
     compile 'org.jboss.resteasy:resteasy-jackson-provider:2.3.3.Final'
     compile 'org.jboss.resteasy:resteasy-multipart-provider:2.3.3.Final'
-    compile 'com.google.guava:guava-base:r03'
```
bruderol: why was that dependency there - not remember we ever used that.

scenarioo-server/src/main/java/org/scenarioo/rest/application/ScenariooWebApplication.java
```diff
@@ -35,41 +35,42 @@
  * Scenarioo REST Services Web Application context that initializes data on startup of server.
  */
 public class ScenariooWebApplication implements ServletContextListener {
-	
+
```
bruderol: I think we have to discuss about our formattings in the code:
1. Each developer should in general make sure to not commit such changes on a feature branch and talk to us to find a better solution
2. Option A: we adjust the formatting of the whole code to better conform to IntelliJ Java default settings and document how to configure same settings in Eclipse
3. Option B: we find a way to configure also IntelliJ to use the same formattings and force every developer to use those.

I vote for option A - let's make our style of code formatting more "common sense" - no extra rules, that most developers do not have in their settings.

Nevertheless, developers should avoid to commit such changes on a feature branch or fork

scenarioo-server/src/main/java/org/scenarioo/model/configuration/Configuration.java
```diff
@@ -54,6 +54,8 @@
 
 	private String scenarioPropertiesInOverview;
 
+	private String elasticsearchEndpoint;
```
bruderol: elasticSearchEndpoint please

scenarioo-server/src/main/java/org/scenarioo/model/configuration/Configuration.java
```diff
@@ -130,6 +132,14 @@ public void setScenarioPropertiesInOverview(final String scenarioPropertiesInOve
 		this.scenarioPropertiesInOverview = scenarioPropertiesInOverview;
 	}
 
+	public String getElasticsearchEndpoint() {
```
bruderol: getElasticSearchEndpoint

scenarioo-server/src/main/java/org/scenarioo/dao/search/IgnoreUseCaseSetStatusMixIn.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package org.scenarioo.dao.search;
+
+import org.codehaus.jackson.annotate.JsonIgnore;
+import org.scenarioo.model.docu.entities.Status;
+
+public abstract class IgnoreUseCaseSetStatusMixIn {
+	@JsonIgnore
+	public abstract void setStatus(Status value);
```
bruderol: this does not have to be solved as a mixed in, just put it directly on the entity, there is no reason to ever use this setter when deserializing, since the one to be used for deserialization has to be the setStatus(String) method.

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		List<SearchDao> searchResults = searchAdapter.searchData(buildIdentifier, q);
+
+		return buildObjectTree(searchResults, q);
+	}
+
+	public void indexUseCases(final UseCaseScenariosList useCaseScenariosList, final BuildIdentifier buildIdentifier) {
+		if(!searchAdapter.isEngineRunning()) {
+			LOGGER.info("No search engine running.");
```
bruderol: this adds a lot of noise to the log to log this allways as info. Please avoid this.
Just do nothing, when the engine is not running.

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	public boolean isEngineRunning() {
+		return searchAdapter.isEngineRunning();
+	}
+
+	public SearchTree search(String q, BuildIdentifier buildIdentifier) {
+		if(!searchAdapter.isEngineRunning()) {
+			LOGGER.info("No search engine running.");
+			return SearchTree.empty();
```
bruderol: Why dont we throw an exception if search is not availbale in that case?
I think it is strange to return an empty collection in that case.

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
```diff
@@ -0,0 +1,115 @@
...
+
+	public void indexSteps(List<Step> steps, List<StepLink> stepLinkList, Scenario scenario, UseCase usecase, BuildIdentifier buildIdentifier) {
+		if(!searchAdapter.isEngineRunning()) {
+			LOGGER.info("No search engine running.");
+			return;
+		}
+
+		searchAdapter.indexSteps(steps, stepLinkList, scenario, usecase, buildIdentifier);
+
+		LOGGER.info("Indexed steps " + buildIdentifier);
```
bruderol: useless log - which steps where indexed ??

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
```diff
@@ -0,0 +1,115 @@
...
+	public void indexUseCases(final UseCaseScenariosList useCaseScenariosList, final BuildIdentifier buildIdentifier) {
+		if(!searchAdapter.isEngineRunning()) {
+			LOGGER.info("No search engine running.");
+			return;
+		}
+
+		searchAdapter.setupNewBuild(buildIdentifier);
+		searchAdapter.indexUseCases(useCaseScenariosList, buildIdentifier);
+
+		LOGGER.info("Indexed build " + buildIdentifier);
```
bruderol: shouldnt it be "Indexed use case " + useCaseScenariosList.getUseCase().getName() ??
Or really needed this log statement?

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	public void indexSteps(List<Step> steps, List<StepLink> stepLinkList, Scenario scenario, UseCase usecase, BuildIdentifier buildIdentifier) {
+		if(!searchAdapter.isEngineRunning()) {
+			LOGGER.info("No search engine running.");
```
bruderol: remove

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
```diff
@@ -0,0 +1,115 @@
...
+		searchAdapter.indexSteps(steps, stepLinkList, scenario, usecase, buildIdentifier);
+
+		LOGGER.info("Indexed steps " + buildIdentifier);
+	}
+
+	private SearchTree buildObjectTree(List<SearchDao> searchResults, final String q) {
+		return new SearchTree(searchResults, q);
+	}
+
+	public void updateAvailableBuilds(List<BuildImportSummary> availableBuilds) {
```
bruderol: I not understand this - what is this for?

scenarioo-server/src/main/java/org/scenarioo/rest/search/SearchResource.java
```diff
@@ -0,0 +1,56 @@
...
+		BuildIdentifier buildIdentifier = ScenarioDocuBuildsManager.INSTANCE.resolveBranchAndBuildAliases(branchName,
+				buildName);
+
+		FullTextSearch search = new FullTextSearch();
+		return search.search(q, buildIdentifier).buildObjectTree();
+	}
+
+	@GET
+	@Path("/searchEngine")
+	public Response isEngineRunning() {
```
bruderol: This special call should not be bound to a branch and build as parameters in the URL (see path on the whole class) that are then not mapped at all,  since it only generally checks the availability of a search engine in general.

Also somehow I think it should simply return false and not an error code if the engine is not there. Status.NOT_FOUND is for resources that are not there, which is something differently. Or if you really think the status code is the better solution, then we should maybe rename it to "searchEngineStatus" instead of "isEngineRunning", which makes it more clear that no boolean is returned.

bruderol: also I think it is somehow not working, at least on the demo it returns OK - but is elastic search really configured on the demo of the fork allready? By the way that is one thing to be resolved then.

bruderol: forget this last comment, I see it now, it is working on the demo. sorry. But the first comment is still important.

scenarioo-server/src/main/java/org/scenarioo/rest/application/ScenariooWebApplication.java
```diff
@@ -35,41 +35,42 @@
...
 	public void contextInitialized(final ServletContextEvent servletContextEvent) {
 		LOGGER.info("====================================================");
 		LOGGER.info("Scenarioo Viewer is starting up ...");
 		LOGGER.info("====================================================");
-		
+
 		initializeApplicationVersion(servletContextEvent.getServletContext());
 		loadConfiguration(servletContextEvent);
-		
+		initializeApplicationContext(servletContextEvent.getServletContext());
```
bruderol: I would either inline this method, or give it a more precise name, like initializeContextPath

scenarioo-server/src/main/java/org/scenarioo/dao/search/IgnoreUseCaseSetStatusMixIn.java
featureenvy: Currently scenarioo-java has no dependency on Jackson, so adding it there for UseCase/Scenario/Step is currently not possible. Is it OK to add the dependency to it? Then we could migrate the mixins directly into the entities.

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
featureenvy: These are all use cases for a given build, so we cannot narrow it down to a single use case for the log. And since this is the main entry point I thought it would be nice to see in the logs that the build really got indexed in case something strange happens.

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
featureenvy: Do you mean the method "buildObjectTree"? I guess by now it is so simple it could be removed.

scenarioo-server/src/main/java/org/scenarioo/rest/search/SearchResource.java
featureenvy: The idea behind the "not found" was to make this call tell if the "search" resource is available or not. If the search engine is not running, the whole resource will never return anything. However we could change this to just return a boolean if you like. And we could bind the call to "rest/searchEngine" (and move the whole branch/{branchName}" etc. from the class to the method in the Java source file, if you think this better aligns with existing Scenarioo resources.

scenarioo-client/app/shared/navigation/navigation.controller.js
adiherzog: This controller was not migrated to vm yet, this is part of https://github.com/scenarioo/scenarioo/issues/509 and does not need to be changed in this PR.

scenarioo-client/app/shared/searchEngineStatus.service.js
adiherzog: I am removing this and am also changing the behavior of the REST service to return a boolean instead of using the HTTP status to indicate whether the service is running or not (see other comment further down).

scenarioo-server/src/main/java/org/scenarioo/rest/search/SearchResource.java
adiherzog: I changed this to not use the status code anymore.

scenarioo-server/src/main/java/org/scenarioo/dao/search/IgnoreUseCaseSetStatusMixIn.java
bruderol: @featureenvy: I still not get it. Why should this setter ever be used?? this field is a pure String field. This special setter with the Status enum as a parameter will never be used when deserializing in my oppinion, it is simply a utility method for convenience, that is not relevant ever when deserializing, because it will use setStatus(String) instead. (sorry for only seeing your comment on that now). Hmm, maybe we have to discuss that once.

scenarioo-client/app/shared/navigation/navigation.controller.js
bruderol: so, we still have unmigrated controllers, I was not aware of that.
Sorry.

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
bruderol: I meant `updateAvailableBuilds` - I guess ...

scenarioo-server/src/main/java/org/scenarioo/dao/search/IgnoreUseCaseSetStatusMixIn.java
bruderol: @adiherzog TODO neuer Kommentar 1, siehe oben 🍻 🍻 🍻

scenarioo-server/src/main/java/org/scenarioo/rest/search/SearchResource.java
bruderol: Thanks @adiherzog - sorry, I didnt see all the answers, because github hides all the outdated diff comments :-(

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
bruderol: @adiherzog TODO Das war der zweite Kommentar dne ich gemeint habe 🍺 🍻 🍻 🍻 
Have a nice weekend!
;-)

scenarioo-server/src/main/java/org/scenarioo/dao/search/FullTextSearch.java
bruderol: ah, ok, allright, that is probably okay. Maybe I just missunderstood. Sorry. Thanks a lot @featureenvy

scenarioo-server/src/main/java/org/scenarioo/dao/search/IgnoreUseCaseSetStatusMixIn.java
adiherzog: @bruderol Yes I verified it, Jackson throws an exception without the MixIn, because it does not know which of the two setters to use when populating the Java object with the data from the JSON message.

featureenvy: That was the problem and why we added the Mixin. We thought about fixing it in the domain object itself in scenarioo-java, but scenarioo-java currently has no dependency on Jackson (at least the last time I checked). If you don't mind the dependency there it would also be a possibility to add the ignore directly on the domain object.

PR #556, opened 2016-10-28
Fulltext Search Improvements

featureenvy:
And another pull request!

Highlights include:
- Better search results thanks to fuzzy search with an "AND" logic instead of term search! This means slight typos (for example "donat" instead of "donate" will still find a result). The AND logic means that all terms in the search query have to match something, otherwise no result is found. I guess this is what a Scenarioo user wants in most cases.
- More search results! Before scenarioo only returned the default number of search results (which is 10). Now it returns 200. It also shows in the search results how many results are shown and how many are found ("showing X of Y hits"). We might want to extend this with a paging functionality in the future (Elasticsearch should support that more or less out of the box). Maybe we should also make the 200 results configurable.
- Add an (very ugly) option to include HTML in the search or not! Yes, it really is that ugly...

### DIFF
--- a/scenarioo-client/app/search/search.html
+++ b/scenarioo-client/app/search/search.html
@@ -25,7 +25,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
      data-treedata="vm.results.resultSet.children"
      data-treemodel="vm.treemodel"
      data-click-action="vm.goToRelatedView(node)"
-     data-first-column-title="Search Results"
+     data-first-column-title="Search Results (showing {{vm.hits}} of {{vm.totalHits}} hits)"
      ng-show="vm.results.resultSet.children.length > 0">
 </div>
 
--- a/scenarioo-client/app/shared/navigation/navigation.html
+++ b/scenarioo-client/app/shared/navigation/navigation.html
@@ -101,6 +101,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                         </button>
                     </span>
                     <input id="sc-global-search-box-textfield" class="form-control input-sm" ng-model="globalSearch.queryString" type="text">
+                    <span class="input-group-addon">
+                        <input type="checkbox" ng-model="globalSearch.includeHtml"> HTML?
+                    </span>
                 </div>
             </form>
         </li>
--- a/scenarioo-server/src/main/java/org/scenarioo/dao/search/dao/SearchResultsDao.java
+++ b/scenarioo-server/src/main/java/org/scenarioo/dao/search/dao/SearchResultsDao.java
@@ -1,0 +1,50 @@
+/* scenarioo-server
+ * Copyright (C) 2014, scenarioo.org Development Team
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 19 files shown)


scenarioo-client/app/search/search.html
```diff
@@ -25,7 +25,7 @@
      data-treedata="vm.results.resultSet.children"
      data-treemodel="vm.treemodel"
      data-click-action="vm.goToRelatedView(node)"
-     data-first-column-title="Search Results"
+     data-first-column-title="Search Results (showing {{vm.hits}} of {{vm.totalHits}} hits)"
```
bruderol: Since we do not have a real paging yet, does displaying "hits of totalHits" make sense at all. Or would we see something like "200 of 15642 hits" in case there are more hits ??

adiherzog:
If will fix the issues that @bruderol found after merging to the develop branch. Merging it now.

bruderol:
:+1: Thanks a lot! also many thanks to @featureenvy ! 👍

scenarioo-server/src/main/java/org/scenarioo/rest/search/SearchResource.java
```diff
@@ -38,14 +31,15 @@
 	@Produces("application/json")
 	@Path("/branch/{branchName}/build/{buildName}/search/{q}")
 	public SearchResponse search(@PathParam("branchName") final String branchName,
-												  @PathParam("buildName") final String buildName, @PathParam("q") final String q) {
+								 @PathParam("buildName") final String buildName, @PathParam("q") final String q,
+								 @QueryParam("includeHtml") boolean includeHtml) {
```
adiherzog: Let's do this when we need it. It's probably not so easy to do this as it's a query parameter at the moment.

scenarioo-client/app/shared/navigation/navigation.html
```diff
@@ -101,6 +101,9 @@
                         </button>
                     </span>
                     <input id="sc-global-search-box-textfield" class="form-control input-sm" ng-model="globalSearch.queryString" type="text">
+                    <span class="input-group-addon">
+                        <input type="checkbox" ng-model="globalSearch.includeHtml"> HTML?
+                    </span>
```
adiherzog: Done, see #565.

scenarioo-server/src/main/java/org/scenarioo/dao/search/dao/SearchResultsDao.java
```diff
@@ -0,0 +1,50 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.scenarioo.dao.search.dao;
+
+import java.util.Collections;
+import java.util.List;
+
+public class SearchResultsDao {
```
adiherzog: I renamed the `Dao`s to `SearchableXyz`, as they are used for indexing and for the search results. Furthermore I renamed the MetaData to `SearchableObjectContext`. I chose this rather long name because it is also used to ignore this part in the index (see `ElasticSearchIndexer`) and I want to avoid that we ignore everything that has "Context" in the name in the index.

scenarioo-client/app/search/search.html
adiherzog: @bruderol Yes, as soon as there are more than 200 results, this is visible here. I tested it with setting the limit to 20:
image

scenarioo-server/src/main/java/org/scenarioo/dao/search/dao/SearchResultsDao.java
bruderol: 👍 great solution! done very well! thanks @adiherzog

scenarioo-client/app/search/search.html
bruderol: alright - Thanks!

Issue #546, opened 2016-10-16
Scenario Overview: Inside an added scenario steps are not reported as "added"


bruderol:
Example see here:
http://demo.scenarioo.org/scenarioo-fork-diffviewer/#/scenario/SwitchLanguageUserInterfaceTest/search_article_in_german_and_switch_to_spanish?branch=wikipedia-docu-example-dev&build=last%20successful&tab=usecases&comparison=To%20last%20successful%20(dev)

This is an added scenario.

Inconsistency: On each step page a step is displayed as "added" but not on the scenario overview page.

danielsuter:
This seems to be the case for all added scenarios. The expectation would an added label be, right?

bruderol:
I am not sure if an "added" label is correct in that case (in my oppinion not optimal as well).
but: at least it would be consistent with how we display it in the step view.
So short answer: yes, unless we define to treat added scenarios differently. 😄

danielsuter:
@bruderol Please verify my PR.

bruderol:
tested and seems to be ok

PR #561, opened 2016-11-07
546-added-steps

danielsuter:
#546 An added label is shown on the scenario overview, if the complete scenario was added.
Step diff data is queried serially, because we don't know if the whole scenario was added before.
Also contains version upgrades for protractor.

### DIFF
--- a/scenarioo-client/app/scenario/scenario.controller.js
+++ b/scenarioo-client/app/scenario/scenario.controller.js
@@ -200,27 +200,67 @@ function ScenarioController($filter, $routeParams,
         return transformMetadataToTree(stepInformation);
     }
 
+    function loadStepDiffInfos(baseBranchName, baseBuildName, comparisonName, pagesAndSteps) {
+        ScenarioDiffInfoResource.get(
+            {
+                'baseBranchName': baseBranchName,
+                'baseBuildName': baseBuildName,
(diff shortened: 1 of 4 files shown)


bruderol:
@danielsuter I do not get it, why this was so difficult to realize, the step page allready displayed the same label before. Is it there solved in the same way (allways having to check for the use case or scenario is an added one)? Shouldnt the service allready return the right diff info for such steps?

danielsuter:
@bruderol The amount of calls stay the same, so I don't really understand your comment.
We know this information in the use case overview page, but not in the scenario overview.

In the step page, the following is done:
1. Load diff info
2. On failure, assume it's added.
This works but results in an error 500 printed on the console, which is not a good solution. We don't have another option there though.

Another difference is, that we only have a step scope on the step overview. Pages don't matter.

bruderol:
ok, allright, I was just surprised. But ok, that looks good to me, as long as it works.

danielsuter:
It would be possible but the solution is not clean as it results in errors on our console. We can't get rid of that one, because it's framework code. Imho my solution is cleaner but it would result in another call on the step page.

bruderol:
ok, let's just leave it like that then. Should I merge or will you do so?

danielsuter:
I'll leave the honour to you :-)

Issue #545, opened 2016-10-16
DiffViewer: GraphicMagick Status Display


bruderol:
Same as for Elastic Search there should be an information about graphicMagick Status on Configuration page with a link to the Documentation about how to install it properly, in case GraphicMagick is currently not available.

danielsuter:
@bruderol 
Please comment on my PR. Architecture is not clear to me yet.

bruderol:
@danielsuter Let's resolve this shortly in pair programming tomorrow - I think that will be easy to "warm up" ;-)

bruderol:
done

PR #563, opened 2016-11-07
545 status display

danielsuter:
Minimalistic implementation of issue #545.

**Why an extra resource?**
Our Configuration domain class is used as configuration.xml file for persistent changes. It is also used in our REST service. If GraphicsMagick is available should not be persisted on the file system but it should be returned in the REST resource. So we could use an extra transfer class there or an additional resource (which implemented). 

I can implement both.

Why are we also returning xml everyhwere? We only consume json. By providing both, we need to support both.

### DIFF
--- a/scenarioo-client/app/manage/generalSettings/generalSettings.controller.js
+++ b/scenarioo-client/app/manage/generalSettings/generalSettings.controller.js
@@ -17,7 +17,7 @@
 
 angular.module('scenarioo.controllers').controller('GeneralSettingsController', GeneralSettingsController);
 
-function GeneralSettingsController($scope, BranchesResource, ConfigService, VersionResource, SearchEngineStatusService, DiffViewerConfigResource) {
+function GeneralSettingsController($scope, BranchesResource, ConfigService, VersionResource, ScenariooStatusService) {
 
     var vm = this;
     vm.branches = [];
@@ -44,12 +44,8 @@ function GeneralSettingsController($scope, BranchesResource, ConfigService, Vers
             }
         );
 
-        DiffViewerConfigResource.get(function onSuccess(result) {
-            vm.diffViewerAvailable = result.graphicsMagickAvailable;
-        });
-
         ConfigService.load();
-        loadSearchEngineStatus();
+        loadServiceStatus();
     }
 
     $scope.$on(ConfigService.CONFIG_LOADED_EVENT, function () {
@@ -69,9 +65,13 @@ function GeneralSettingsController($scope, BranchesResource, ConfigService, Vers
         }
     }
 
-    function loadSearchEngineStatus () {
-        SearchEngineStatusService.isSearchEngineRunning().then(function(result) {
+    function loadServiceStatus () {
+        ScenariooStatusService.isSearchEngineRunning().then(function(result) {
             vm.searchEngineStatus = result;
+        });
+
+        ScenariooStatusService.isDiffViewerImageProcessingAvailable().then(function(result) {
+            vm.diffViewerAvailable = result.graphicsMagickAvailable;
         });
     }
 
--- a/scenarioo-client/app/manage/generalSettings/generalSettings.controller.js
+++ b/scenarioo-client/app/manage/generalSettings/generalSettings.controller.js
@@ -17,7 +17,7 @@
 
 angular.module('scenarioo.controllers').controller('GeneralSettingsController', GeneralSettingsController);
 
-function GeneralSettingsController($scope, BranchesResource, ConfigService, VersionResource, ScenariooStatusService) {
+function GeneralSettingsController(BranchesResource, ConfigService, ApplicationStatusService) {
 
     var vm = this;
     vm.branches = [];
@@ -38,20 +38,18 @@ function GeneralSettingsController($scope, BranchesResource, ConfigService, Vers
             calculateConfiguredBranch();
         });
 
-        VersionResource.get(
-            function onSuccess(result) {
-                vm.version = result;
-            }
-        );
+        ApplicationStatusService.getApplicationStatus().then(function(status) {
 
-        ConfigService.load();
-        loadServiceStatus();
+            vm.version = status.version;
+            vm.configuration = status.configuration;
+            vm.diffViewerAvailable = status.diffViewerStatus.isGraphicsMagickAvailable;
+            vm.searchEngineStatus = status.searchEngineStatus;
+
+            calculateConfiguredBranch();
+
+        });
+
     }
-
-    $scope.$on(ConfigService.CONFIG_LOADED_EVENT, function () {
-        vm.configuration = ConfigService.getRawConfigDataCopy();
-        calculateConfiguredBranch();
-    });
 
     function calculateConfiguredBranch() {
         if (angular.isUndefined(vm.branches) || angular.isUndefined(vm.configuration)) {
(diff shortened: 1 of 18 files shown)


bruderol:
providing XML and JSON both, comes for free and is very easy. But I thought Adrian removed the XML support once? Allthough I liked it, because it is sometimes nicer to read for debugging. but ok you can remove, allthough I thought it is nice to have a REST endpoint that supports both.

danielsuter:
It does not matter at me for the moment, so I won't remove it.
Did you also have time to look at my other questions?

scenarioo-client/app/manage/generalSettings/generalSettings.controller.js
```diff
@@ -43,6 +44,10 @@ function GeneralSettingsController($scope, BranchesResource, ConfigService, Vers
             }
         );
 
+        DiffViewerConfigResource.get(function onSuccess(result) {
```
bruderol: I propose to refactor DiffViewerConfigResource and put it together with SearchEngineStatusService and have instead one endpoint to get all these statuses at once. We could simply call it ScenariooStatusResource (also the REST-endpoint should become the same). @danielsuter waht you think? I understand the argument, that it belongs not to the config endpoint.

scenarioo-client/app/manage/generalSettings/generalSettings.controller.js
```diff
@@ -43,6 +44,10 @@ function GeneralSettingsController($scope, BranchesResource, ConfigService, Vers
...
+            vm.diffViewerAvailable = result.graphicsMagickAvailable;
+        });
+
         ConfigService.load();
         loadSearchEngineStatus();
```
bruderol: put the other status loading code together with other status code into this private mehtod here and rename it to `loadScenariooStatus`

scenarioo-client/app/manage/generalSettings/generalSettings.controller.js
bruderol: one other option could also be, to even refactor the config endpoint to deliver all this state (configuration object and all other status info) in one call, But this could still be done, when really needed. Two calls are allright for me.

bruderol:
@danielsuter see my comments. Does this answer your questions? I think one REST endpoint for all such status information would make sense (now we have two for full text search and for diff viewer and both are loaded in same place). In my oppinion later even the config could then also become part of this new endpoint, but for now we could leave it like that.

scenarioo-client/app/manage/generalSettings/generalSettings.html
```diff
@@ -54,12 +54,27 @@
...
+            </div>
+        </div>
+
         <h2>Full Text Search</h2>
 
         <div id="sc-search-not-configured-message" class="row sc-space-bottom" ng-show="vm.searchEngineStatus && !vm.searchEngineStatus.endpointConfigured" ng-cloak>
             <div class="col-lg-12">
-                Full text search is disabled. <br/>
-                See <a href="https://github.com/scenarioo/scenarioo/blob/{{vm.version.releaseBranch}}/docs/features/full-text-search/setup.md" target="_blank">Setup Instructions for Full Text Search</a> on how to configure and setup Full text search for Scenarioo.
+                Full text search is <span class="sc-service-disabled">disabled</span>. <br/>
```
danielsuter: There's no message for enabled, just input fields. That's why I couldn't format it the same.

scenarioo-client/app/manage/generalSettings/generalSettings.html
```diff
@@ -54,12 +54,27 @@
             </div>
         </div>
 
+        <h2>Diff viewer</h2>
+
+        <div id="sc-diff-viewer-not-configured-message" class="row sc-space-bottom" ng-show="!vm.diffViewerAvailable" ng-cloak>
+            <div class="col-lg-12">
+                Image comparison is <span class="sc-service-disabled">disabled</span>. Structural comparisons are enabled.<br/>
+                See <a href="https://github.com/scenarioo/scenarioo/blob/{{vm.version.releaseBranch}}/docs/features/diff-viewer/setup.md" target="_blank">Setup instructions for Diff Viewer</a> on how to configure and setup Diff Viewer for Scenarioo.
```
danielsuter: Yep, that's great!

scenarioo-server/src/main/java/org/scenarioo/rest/configuration/ConfigurationResource.java
```diff
@@ -41,7 +42,16 @@
...
+
+	@GET
+	@Path("/diffViewer")
+	@Produces({ "application/json", "application/xml" })
+	public DiffViewerStatus getDiffViewerStatus() {
+		DiffViewerStatus status = new DiffViewerStatus();
+		status.setGraphicsMagickAvailable(GraphicsMagickConfiguration.isAvailable());
+		return status;
+	}
+
```
danielsuter: That's the catch, there's none. Fulltext data is persisted into our configuration.

danielsuter:
@bruderol 
The problem is, that elastic search information is actually persisted into our configuration. That's why I can't handle it the same or I think it's not the same. Is this probably something for our coding day?

bruderol:
persisted into our configuration?? Really ?? But it looks like loading it from an extra endpoint (similar to you).

danielsuter:
hmm let me recheck...

bruderol:
ähm, nope, we are talking about two different things ... yes, there isa config for the nedpoint, right, but the status, whether the endpoint is reachable or not id loaded from a special REST endpoint - see `SearchResource` and `SearchEngineStatusResponse`. My proposal is to refactor this into one nice `ScenariooStatusResource` and just do one additional call on the config page (now there are 3).

danielsuter:
You're right of course. I misunderstood you. I'll refactor as suggested. That was the solution I was looking for!  :thumbsup:

bruderol:
@danielsuter see `generalSettings controller : loadSearchEngineStatus()`
but you will have to move the one liner from the SearchResource to the new status resource.

bruderol:
@danielsuter ev. musst du aufpassen, weil der Search Status häufiger abgefragt wird ... ev. auch noch an anderen Stellen. trotzdem denke ich, dass man zumindest für die Konfig Seite alles in einen endpoint zusammenpacken und an den client senden kann, kannst ja den separaten Status für die Suche noch stehen lassen, für andere Verwendungszwecke.

danielsuter:
@bruderol 
Eventuell verstehe ich es noch immer nicht ganz. Sonst müssen wir das am Coding Day noch fertig anschauen. Hab das mal in einen eigenen Service gepackt.

bruderol:
@danielsuter Ich meine einen Call "getScenariooStatus" welcher ein Objekt "ScenariooStatus" zurück liefert und somit mit einem Server-Roundtrip alle Stati zurück liefert, welche auf der Config-Seite angezeigt werden sollen.

scenarioo-client/app/manage/generalSettings/generalSettings.controller.js
```diff
@@ -64,10 +65,14 @@ function GeneralSettingsController($scope, BranchesResource, ConfigService, Vers
         }
     }
 
-    function loadSearchEngineStatus () {
-        SearchEngineStatusService.isSearchEngineRunning().then(function(result) {
+    function loadServiceStatus () {
+        ScenariooStatusService.isSearchEngineRunning().then(function(result) {
```
bruderol: this call and the other should be one, that returns an object !

scenarioo-server/src/main/java/org/scenarioo/rest/configuration/ConfigurationResource.java
```diff
@@ -41,7 +42,16 @@
 	public Configuration getConfiguration() {
 		return configurationRepository.getConfiguration();
 	}
-	
+
+	@GET
+	@Path("/diffViewer")
+	@Produces({ "application/json", "application/xml" })
+	public DiffViewerStatus getDiffViewerStatus() {
```
bruderol: @danielsuter What i mean: this service should return a new class called ScenariooStatus, and this status should contain all kind of stati that is needed in config page (I would even include the Configuration object as well), so we just have to call one service for everything needed on this config page.

The service would be called "getScenariooStatus" then and it will also return the search engine status in the same call.

bruderol:
thanks a lot @danielsuter ! 👍

Issue #547, opened 2016-10-16
Step Comparison: Legend for highlighted changes


bruderol:
Legend "highlighted changes" should be only displayed ...
1. only over the screen that realy has highlighted changes (not over both!)
2. only if "show highlights" is turned on

![image](https://cloud.githubusercontent.com/assets/755721/19421273/33a15c28-93fd-11e6-96a2-c5e38cbb6ff9.png)

danielsuter:
Please review.

bruderol:
done, tested, seems to be good 👍 Thanks!

PR #564, opened 2016-11-07
#547 Only show highlights when needed.

danielsuter:
#547
Fixed directive.
Refactored template, so the title is properly aligned.

### DIFF
--- a/scenarioo-client/app/step/step.html
+++ b/scenarioo-client/app/step/step.html
@@ -144,26 +144,35 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                     </div>
                 </div>
                 <div id="sc-step-comparison-side-by-side-view" ng-if="comparisonViewOptions[displaySideBySideViewKey]">
-                    <div class="col-lg-6">
-                        {{baseBuild.aliasName}}
-                        <div class="sc-base-screenshot clear" ng-if="!comparisonViewOptions[displayDiffScreenshotKey]">
-                            <sc-screenshot-title build="baseBuild" name="Current" alias="{{baseBuildName}}" highlights="false"></sc-screenshot-title>
-                            <sc-annotated-screenshot screen-shot-url="screenShotUrl" visibility-toggle="showScreenAnnotations"></sc-annotated-screenshot>
+                    <div class="row">
+                        <div class="col-lg-6">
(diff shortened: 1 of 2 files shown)


scenarioo-client/app/step/step.html
```diff
@@ -144,26 +144,35 @@
                     </div>
                 </div>
                 <div id="sc-step-comparison-side-by-side-view" ng-if="comparisonViewOptions[displaySideBySideViewKey]">
-                    <div class="col-lg-6">
-                        {{baseBuild.aliasName}}
-                        <div class="sc-base-screenshot clear" ng-if="!comparisonViewOptions[displayDiffScreenshotKey]">
-                            <sc-screenshot-title build="baseBuild" name="Current" alias="{{baseBuildName}}" highlights="false"></sc-screenshot-title>
-                            <sc-annotated-screenshot screen-shot-url="screenShotUrl" visibility-toggle="showScreenAnnotations"></sc-annotated-screenshot>
+                    <div class="row">
```
bruderol: having two rows layout here seems risky to me ... for small screens it may mean that there will be the headers first and then both screenshots. No one will understand that.

scenarioo-client/app/step/step.html
```diff
@@ -144,26 +144,35 @@
...
                         </div>
                     </div>
-                    <div class="col-lg-6">
-                        <div>
-                            <sc-screenshot-title build="comparisonBuild" name="{{comparisonName}}" highlights="false"></sc-screenshot-title>
-                            <div ng-if="step.diffInfo.isAdded">
-                                <p>No comparison screenshot because this is an added step.</p>
+                    <div class="row">
+                        <div class="col-lg-6">
+                            {{baseBuild.aliasName}}
```
bruderol: if the build has an alias name, I think that this should be displayed as part of the title row, or not? Before this was even in front of the rest of the title, which I think was good, if I remember correctly.

bruderol:
@danielsuter we could merge and then test it on the demo and then still correct .. waht do you think? see my comments - I amcurious to see it in action!

scenarioo-client/app/step/step.html
danielsuter: hmm true, otherwise the title will miss align, which looks bad on large screen, where Scenarioo is mostly used.

PR #578, opened 2016-12-24
Several Proposed DiffViewer Changes

bruderol:
We have to check whether we can and want to merge this changes from a fork and want to merge that as it is ... I will coordinate and discuss with them what of this we can use as is and what needs to be ajusted, we will see ...

@taascma1 @til77 Thanks for your email - I will review and contact you for further discussions but maybe only next year! happy holidays!


bruderol:
Problem no 1: this commits seem to be made with a not properly configured github username/account ... You should correct that please!

@til77 ??

bruderol:
Since the fork has changed git history - I close this pull request and opened a new one here:
#593

Issue #591, opened 2017-02-22
Epic: DiffViewer Plus Improvements


bruderol:
Review and merge changes from fork https://github.com/til77/scenarioo

Pull Request for reviewing:
https://github.com/scenarioo/scenarioo/pull/593

Let's collect the issues that should be addressed here in this epic:

* [x] #601 - ordering of branches in branch selection list.
* [x] #596 - default highlight color for diff screenshots configurable
* [x] #577 - no confusing null pointer exceptions that should not be there (assumed to be fixed)
* [x] #602 - Remove GraphicMagick Dependency
* [x] #618 - new internal diff viewer format (less disk space, attached to belonging build)
* [x] #603 - Review and adjust Step Comparison View

further issues - done later ... (see Milestone https://github.com/scenarioo/scenarioo/milestone/33 )
* [ ] #605 - Change and document new REST API for diff calculation
* [ ] #600 - avoid server blocked from startup if there are build data inconsistencies (to be verified!).
* [ ] #598 - avoid further problems with outdated libraries in java 8
* [ ] #604 - improve background comparison calculation (use power of multi threading)

Out Of Scope (no time):
* #573 - no disabled comparison in URL
* #568 - responsive layout of labels in step comparison view for small screens
* #599 - advanced views to see change rate for each page as well and all screens of one page.

kabl:
In Review:
- #601 - ordering of branches in branch selection list.
- #596 - default highlight color for diff screenshots configurable

bruderol:
I will close this, see remaining issues for what happens to them.

All reamining work for the realse is here on the milestone and will be tracked there:
https://github.com/scenarioo/scenarioo/milestone/33

PR #593, opened 2017-02-22
Feature/591 Diff Viewer Improvements

bruderol:
Proposed changes by @til77 to be reviewed and further discussed how to deal and proceed with them.

Demo for this feature branch can be found here: http://demo.scenarioo.org/scenarioo-diff-viewer-plus 

See also issue #591


scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
```diff
@@ -18,167 +18,149 @@
...
+			// convert images to ARGB format to ensure that the data buffers store one pixel as ARGB
+			BufferedImage oldImageArgb = new BufferedImage(unifiedWidth, unifiedHeight, BufferedImage.TYPE_INT_ARGB);
+			BufferedImage newImageArgb = new BufferedImage(unifiedWidth, unifiedHeight, BufferedImage.TYPE_INT_ARGB);
+
+			// for simplicity paint both images on the upper left corner of the compare image since searching the sub-image is too expensive
+			// areas not covered by images are transparent which will lead to a maximum difference on all 3 measured color channels
+			oldImageArgb.getGraphics().drawImage(oldImage, 0, 0, null);
+			newImageArgb.getGraphics().drawImage(newImage, 0, 0, null);
+
+			BufferedImage diffImage = new BufferedImage(Math.max(oldImageArgb.getWidth(), newImageArgb.getWidth()), Math.max(oldImageArgb.getHeight(), oldImageArgb.getHeight()), BufferedImage.TYPE_INT_ARGB);
```
kabl: Why oldImage.getWidth() should be different than oldImageArgb.getWidth()? Image size calculation already done in line 97.

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
```diff
@@ -18,167 +18,149 @@
...
+			double absoluteErrorTotal = 0;
+
+			// improve performance by accessing data buffers directly since methods to access pixels through BufferedImage are passed through the ColorModel which slows things down
+			DataBuffer oldImageBuffer = oldImageArgb.getRaster().getDataBuffer();
+			DataBuffer newImageBuffer = newImageArgb.getRaster().getDataBuffer();
+			DataBuffer diffImageBuffer = diffImage.getRaster().getDataBuffer();
+
+			int totalPixels = oldImageBuffer.getSize();
+
+			for (int n = 0; n < oldImageBuffer.getSize(); n++) {
```
kabl: use "totalPixels"

kabl: Maybe could use Java 8 stream api. Doing this work with parallelStream could improve the performance.

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
```diff
@@ -18,167 +18,149 @@
...
-	}
-
 	/**
-	 * A Diff Screenshoot will be created and stored in the directory of the diffScreenshot path.
+	 * A Diff Screenshot will be created and stored in the directory of the diffScreenshot path.
 	 * This directory, and all the parent directories, will be created if they do not exist.
 	 */
-	double compareScreenshots(final File baseScreenshot, final File comparisonScreenshot,
-			final File diffScreenshot) {
+	double compareScreenshots(final File baseScreenshot,
```
kabl: This method is doing more things than compoareScreenshots. It's more about compareCreateRequiredDirectoriesForScreenshotAndStore method. Some Clean Code may help

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
```diff
@@ -18,167 +18,149 @@
...
-			if (m.find()) {
-				return Math.sqrt(Double.parseDouble(m.group(0))) * 100;
+
+			double mae = absoluteErrorTotal / totalPixels;
+
+			if (diffAvailable) {
+				ImageIO.write(diffImage, "PNG", diffScreenshot);
 			}
+
+			return Math.sqrt(mae) * 100;
```
kabl: Magic code

scenarioo-server/src/main/java/org/scenarioo/model/diffViewer/ComparisonResult.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.FIELD)
+public class ComparisonResult {
+
+	private BuildIdentifier baseBuild;
+	private BuildIdentifier compareBuild;
+
+	ComparisonConfiguration comparisonConfiguration;
```
kabl: Why there is no private modifier?

scenarioo-server/src/main/java/org/scenarioo/model/diffViewer/ComparisonResult.java
```diff
@@ -0,0 +1,69 @@
...
+
+	public void setComparisonConfiguration(ComparisonConfiguration comparisonConfiguration) {
+		this.comparisonConfiguration = comparisonConfiguration;
+	}
+
+	public String getRmaePercentage() {
+		return rmaePercentage;
+	}
+
+	public void setRmaePercentage(double rmaePercentage) {
```
kabl: This should be a string. Because getter and setter of a member should be the same type.

scenarioo-server/src/main/java/org/scenarioo/rest/configuration/DiffViewerStatus.java
```diff
@@ -4,13 +4,4 @@
 
 @XmlRootElement
 public class DiffViewerStatus {
```
kabl: Empty class. Should be deleted

scenarioo-server/src/main/java/org/scenarioo/rest/configuration/ConfigurationResource.java
```diff
@@ -52,7 +51,6 @@ public ApplicationStatus getApplicationStatus() {
 		ApplicationStatus applicationStatus = new ApplicationStatus();
```
kabl: This class has Constructor and getter/setter which are never used.

scenarioo-server/src/main/java/org/scenarioo/rest/builds/BuildsImporterResource.java
```diff
@@ -98,6 +95,35 @@ public void reimportBuild(@PathParam("branchName") final String branchName,
 		ScenarioDocuBuildsManager.INSTANCE.reimportBuild(buildIdentifier);
 	}
 
+	@GET
+	@Path("importBuild/{branchName}/{buildName}/{comparisonBranchName}/{comparisonBuildName}/{comparisonName}")
+	@Produces({"application/json"})
+	public Response importBuild(
```
kabl: No tests available

kabl:
Not sure but I guess this documentation needs an update: https://github.com/scenarioo/scenarioo/blob/develop/docs/features/diff-viewer/setup.md

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
bruderol: Indeed - magic - Think we should remove that - @til77 what do you think?

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
bruderol: Cool Idea - Open an issue for it - if you have time you could do that.

bruderol: @kabl not understand your comment?

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
bruderol: yes that looks strange - think we can clean that up

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
bruderol: Maybe not the best name, agree, but your name is not a better name for me - sorry - the proposed alternative would be less understandable for me to quickly understand what this method is doing - maybe something like `compareScreenshotAndStoreDiffImage` - creating parent directories for storing something is something we always do, whenever it is needed for storing something. So that is not realy an unexpected side effect here.

bruderol: by the way - also it is somehow clear from the context, that all the methods here just store their results in the file system, that is why I think the original name is not so bad, as you may think.

@danielsuter What do you think?

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
```diff
@@ -18,167 +18,149 @@
...
+		} catch (IOException ex) {
+			LOGGER.warn("Failed to compare images (base: " + baseScreenshot + ", comparison: " + comparisonScreenshot + ")", ex);
+			return SCREENSHOT_DEFAULT_CHANGE_RATE;
```
bruderol: no - I think not - can explain it to you.

Short answer: Since this is a background job the log is stored in a log file, that the user could browse (either in file system on the server or later through a special view in the web app, which is currently missing --> there is an issue for that)

scenarioo-server/src/main/java/org/scenarioo/model/diffViewer/ComparisonResult.java
bruderol: Dont know, is it used somewhere? Maybe just lazy code?

scenarioo-server/src/main/java/org/scenarioo/rest/configuration/DiffViewerStatus.java
bruderol: yes, indeed - well spotted! --> that will also have impact on the client side, because there is some config view where this status is displayed - that is dead code that has to be removed everywhere it is still used (in REST service and as well on the client side).

scenarioo-server/src/main/java/org/scenarioo/rest/configuration/ConfigurationResource.java
bruderol: which class?

bruderol:
@kabl: concerning last comment: yes, sure - there is an issue for that.

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
bruderol: ah, yes!

scenarioo-server/src/main/java/org/scenarioo/model/diffViewer/ComparisonResult.java
bruderol: why is it a string? why dont we store that as a number? @til77 can you clarify on that?

scenarioo-server/src/main/java/org/scenarioo/rest/builds/BuildsImporterResource.java
bruderol: oh - i see - that one looks like we need to discuss it further - I not like the REST service to wait for potentially endless unless some background tasks are done. We have to discuss a different solution here. @til77  wouldnt it be also okay for you if we simply provide another service that checks for a comparison result and returns nothing if it is not available yet and then let the build poll for that information until it is available?

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
kabl: My method name was not a proposal. It's more about split this method in smaller parts. Or move some calculation in methods.

scenarioo-server/src/main/java/org/scenarioo/rest/configuration/ConfigurationResource.java
kabl: Sorry. The class was not modified in this PR. It's ```ApplicationStatus.java```

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
til77: yes, can be cleaned up by using unifiedWidth and unifiedHeight

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
til77: I have no idea why the sqrt is used. Maybe it was used to emphasize the difference on very small numbers like a diff of 0.0001 to reach one percent change. Our change to always round differences up to 1% maybe makes this calculation obsolete?

scenarioo-server/src/main/java/org/scenarioo/model/diffViewer/ComparisonResult.java
til77: I assume that this was missed.

scenarioo-server/src/main/java/org/scenarioo/model/diffViewer/ComparisonResult.java
til77: I see no reason for the type mess. Change it to double if the xml libraries have no trouble with double.

scenarioo-server/src/main/java/org/scenarioo/rest/builds/BuildsImporterResource.java
til77: Changing to async is also fine for us however the importBuild method would have to return a reference ID so that the client can ask about the status for this particular import.

bruderol: @til77 yes, sure, there is already such an "id" and this is implicitly given by "branch name" + "build name" + the configured "comparison name" - and there is a also a REST service to get the result once it is calculated. Would that be enough?

bruderol: @til77 the URL to get the diff info for a comparison, is: http://demo.scenarioo.org/scenarioo-diff-viewer-plus/rest/diffViewer/baseBranchName/{put-your-branch-name-here}/baseBuildName/{put-your-build-name-here}/comparisonName/{put-the-name-of-the-comparison-here}/buildDiffInfo  -returned json contains the result in field "changeRate" as a decimal number

bruderol: @kabl @til77 I specified in issue #605 what I think we will have to change.

scenarioo-docu-generation-example/.gitignore
```diff
@@ -1,4 +1,5 @@
 build
+out
```
bruderol: @adiherzog Why is this needed now?? I not understand it

scenarioo-docu-generation-example/src/test/resources/example/documentation/scenarioDocuExample/wikipedia-docu-example-dev/2014-04-19/build.xml
```diff
@@ -3,5 +3,6 @@
     <name>2014-04-19</name>
     <revision>F398DA3</revision>
     <date>2014-04-19T00:00:00+02:00</date>
+    <status>success</status>
```
bruderol: @adiherzog This commit to change the test data was unnecessary - the example documentaiton data should always be commited (only!) after a clean run of "scenarioo-docu-generation-example" - this fields are added by the viewer after importing the builds the first time, and it is intended to not have them in the example data for testing that the state gets automatically added by the server. But yes, I know, this is not optimal and we might have to improve that once. until we improve it we should only commit example data, when generated with a clean build.

bruderol: @adiherzog hmmm - ok, forget - it seems the tests are now generating that status for the builds always??? Why??? I think this has been changed wrongly.

bruderol: @adiherzog: I do not get it - somehwo it seems that you copied it from the "out" directory - there it is generated with the state - I not get why there is this "out"-directory - cant we change that to write to build as well and use meaningful name why this output is produced (probably testing the import)

bruderol: @adiherzog ahhhrgh - okay - think now I understand - sorry - my first comment was right - the version in the out directory is always the commited one (gets copied during build somehow) - will now copy again the version from build and then it should be okay again.

scenarioo-docu-generation-example/.gitignore
adiherzog: I had a wrong IntelliJ configuration that compiled into the out folder. So I will remove "out" again.

scenarioo-docu-generation-example/.gitignore
```diff
@@ -1,5 +1,4 @@
 build
-out
```
bruderol: hmm, sorry, maybe my comment about this was bad, I have seen that I have this 'out' as well - so just removing it from gitignore does not realy solve the problem - either put it back or adjust that everything is built into `build` directory

bruderol:
@adiherzog and me decided to merge that now -  we will continue to improve diff viewer further on additional smaller feature branches

PR #594, opened 2017-02-25
Fix syntax inconsistencies in README.md

sne:
- Either use "and" or "&", but do it consistently
- Don't randomly capitalize words
- Don't use "even" so often

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,24 +6,24 @@ develop [![Continuous Build Develop Branch](http://build.scenarioo.org/jenkins/j
 
 ## Automated Documentation of Applications with UI Tests
 
-Scenarioo is a tool to leverage the power of your UI Tests and to make them available as a documentation of your software system to all people involved in the design, development, test, operation and management of your software system.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,24 +6,24 @@ develop [![Continuous Build Develop Branch](http://build.scenarioo.org/jenkins/j
 
 ## Automated Documentation of Applications with UI Tests
 
-Scenarioo is a tool to leverage the power of your UI Tests and to make them available as a documentation of your software system to all people involved in the design, development, test, operation and management of your software system.
+Scenarioo is a tool to leverage the power of your UI tests and to make them available as a documentation of your software system to all people involved in the design, development, test, operation and management of your software system.
 
-### Webpage and Documentation
+### Webpage & Documentation
```
bruderol: take care - changing titles always means that links to here could be broken. Here in this case I think it is not a problem, but should be checked for every title change.

bruderol:
Thanks @sne - very good input - I like the changes - it's indeed slightly better. thanks a lot!

Issue #596, opened 2017-03-07
DiffViewer Improvements - Color of Diff Image


patrickschweizer:
CR / Feedback from one team using Scenarioo:
Make color of Diff Image configurable

Depending on the background or foreground colors used in a web application, a suitable Diff Image color  can be configured so that the contrast can be optimised to highlight differences.

bruderol:
also the ratio of transparency should be configurable.

the default should be stronger than what we had as old default but a little more transparent and orange than in the current changed implementation on the PR.

@adiherzog @danielsuter this will probably be implemented by @kabl

kabl:
start working here...

kabl:
Merge Request: https://github.com/scenarioo/scenarioo/pull/608

kabl:
@bruderol @patrickschweizer: The new configuration value. Example: ```<diffImageColorRgbaHex>0xc8ff0000</diffImageColorRgbaHex>```
Pattern: 0xAARRGGBB.

In some Sub-Classes of the configuration there also exists Color Properties. They're in the HTML RGB Syntax ```#RRGGBB``` but HTML doesn't support an Alpha Channel.

bruderol:
yes, that looks ok for me. Use whatever fits best for the purpose you need it.

kabl:
PR: https://github.com/scenarioo/scenarioo/pull/609

bruderol:
@adiherzog I think the default should be orange again and more transparent let's say 70% or even less. and also it might be worth to have a look at my last comments on the PR.

adiherzog:
Use stronger orange

bruderol:
I changed it to be the same color as used in other places for changes, and changed the color everywhere to a stronger orange (more red)

bruderol:
I close this, it is now configurable, and the default orange for highlights is slightly stronger than it was before.

PR #609, opened 2017-04-24
Feature/596 color of diff image2

kabl:
Issue: https://github.com/scenarioo/scenarioo/issues/596

### DIFF
--- a/scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/AbstractComparator.java
+++ b/scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/AbstractComparator.java
@@ -36,24 +36,24 @@ public abstract class AbstractComparator {
 	protected static final ConfigurationRepository configurationRepository = RepositoryLocator.INSTANCE
 		.getConfigurationRepository();
 
-	protected ComparatorParameter parameter;
+	protected ComparisonParameters parameters;
 	protected ScenarioDocuReader docuReader;
 	protected DiffWriter diffWriter;
 	protected String baseBranchName;
 	protected String baseBuildName;
 	protected ComparisonConfiguration comparisonConfiguration;
 
-	public AbstractComparator(ComparatorParameter parameter) {
-		this.parameter = parameter;
-		this.baseBranchName = parameter.getBaseBranchName();
-		this.baseBuildName = parameter.getBaseBuildName();
-		this.comparisonConfiguration = parameter.getComparisonConfiguration();
+	public AbstractComparator(ComparisonParameters parameters) {
+		this.parameters = parameters;
+		this.baseBranchName = parameters.getBaseBranchName();
+		this.baseBuildName = parameters.getBaseBuildName();
+		this.comparisonConfiguration = parameters.getComparisonConfiguration();
 
 		this.docuReader = new ScenarioDocuReader(
 			configurationRepository.getDocumentationDataDirectory());
-		this.diffWriter = new DiffWriterXmlImpl(parameter.getBaseBranchName(),
-			parameter.getBaseBuildName(),
-			parameter.getComparisonConfiguration().getName());
+		this.diffWriter = new DiffWriterXmlImpl(parameters.getBaseBranchName(),
+			parameters.getBaseBuildName(),
+			parameters.getComparisonConfiguration().getName());
 	}
 
 }
--- a/scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/AbstractComparator.java
+++ b/scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/AbstractComparator.java
@@ -36,21 +36,24 @@ public abstract class AbstractComparator {
 	protected static final ConfigurationRepository configurationRepository = RepositoryLocator.INSTANCE
 		.getConfigurationRepository();
 
+	protected ComporatorParameter parameter;
 	protected ScenarioDocuReader docuReader;
 	protected DiffWriter diffWriter;
 	protected String baseBranchName;
 	protected String baseBuildName;
 	protected ComparisonConfiguration comparisonConfiguration;
 
-	public AbstractComparator(String baseBranchName, String baseBuildName, final ComparisonConfiguration comparisonConfiguration) {
+	public AbstractComparator(ComporatorParameter parameter) {
+		this.parameter = parameter;
+		this.baseBranchName = parameter.getBaseBranchName();
+		this.baseBuildName = parameter.getBaseBuildName();
+		this.comparisonConfiguration = parameter.getComparisonConfiguration();
+
 		this.docuReader = new ScenarioDocuReader(
 			configurationRepository.getDocumentationDataDirectory());
-		this.diffWriter = new DiffWriterXmlImpl(baseBranchName,
-			baseBuildName,
-			comparisonConfiguration.getName());
-		this.baseBranchName = baseBranchName;
-		this.baseBuildName = baseBuildName;
-		this.comparisonConfiguration = comparisonConfiguration;
+		this.diffWriter = new DiffWriterXmlImpl(parameter.getBaseBranchName(),
+			parameter.getBaseBuildName(),
+			parameter.getComparisonConfiguration().getName());
 	}
 
 }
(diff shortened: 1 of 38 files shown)


kabl:
@bruderol ready for Review. Which should be the default color?

bruderol:
@kabl Default Color. Would be cool if we can take some dark strong orange and make it 50% transparent such, that on white backgrounds it looks similar to that orange here: 
image

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ComparatorParameter.java
```diff
@@ -0,0 +1,38 @@
...
+		return baseBuildName;
+	}
+
+	public ComparisonConfiguration getComparisonConfiguration() {
+		return comparisonConfiguration;
+	}
+
+	public Color getDiffImageColor() {
+		return diffImageColor;
+		//return  new Color(255, 0, 0, 200);
```
bruderol: please no dead comments

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/AbstractComparator.java
```diff
@@ -36,21 +36,24 @@
...
 	protected String baseBranchName;
 	protected String baseBuildName;
 	protected ComparisonConfiguration comparisonConfiguration;
 
-	public AbstractComparator(String baseBranchName, String baseBuildName, final ComparisonConfiguration comparisonConfiguration) {
+	public AbstractComparator(ComparatorParameter parameter) {
+		this.parameter = parameter;
+		this.baseBranchName = parameter.getBaseBranchName();
+		this.baseBuildName = parameter.getBaseBuildName();
+		this.comparisonConfiguration = parameter.getComparisonConfiguration();
```
bruderol: hmm, I dont know, now the parameters are now somehow stored in two places ... once in `parameter` and once in the other fields of this class - this seems to me confusing.

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ComparatorParameter.java
```diff
@@ -0,0 +1,38 @@
+package org.scenarioo.business.diffViewer.comparator;
+
+import org.scenarioo.model.configuration.ComparisonConfiguration;
+
+import java.awt.*;
+
+public class ComparatorParameter {
```
bruderol: I think this should be called `ComparisonParameters` cause it contains all parameters for one comparison.

scenarioo-server/src/main/java/org/scenarioo/model/configuration/Configuration.java
```diff
@@ -17,6 +17,8 @@
 
 package org.scenarioo.model.configuration;
 
+import java.awt.*;
```
bruderol: Why dont we just import only what we need?

scenarioo-server/src/main/java/org/scenarioo/model/configuration/Configuration.java
```diff
@@ -226,4 +240,13 @@ public void setCreateLastSuccessfulScenarioBuild(final boolean createLastSuccess
 		this.createLastSuccessfulScenarioBuild = createLastSuccessfulScenarioBuild;
 	}
 
+	public Color getDiffImageColor() {
+		//Required BigInteger. Because the hex value could be out of range (int is signed). With BigInteger we can avoid this issue.
+		int rgba = new BigInteger(diffImageColorRgbaHex.substring(2), 16).intValue();
```
bruderol: Is there no proper way to decode that (by also taking `0x` into account) - like for example Integer.decode is doing.

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
```diff
@@ -39,9 +37,8 @@
 	private static final int SCREENSHOT_DEFAULT_CHANGE_RATE = 0;
 	private DiffReader diffReader;
 
-	public ScreenshotComparator(final String baseBranchName, final String baseBuildName,
-								final ComparisonConfiguration comparisonConfiguration) {
-		super(baseBranchName, baseBuildName, comparisonConfiguration);
+	public ScreenshotComparator(ComparatorParameter parameter) {
```
bruderol: `parameter` should be renamed to `parameters`everywhere since it is a collection of several parameters.

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
```diff
@@ -88,7 +85,8 @@ public double compare(final String baseUseCaseName, final String baseScenarioNam
 		}
 
 		try {
-			int diffColor = new Color(255, 0, 0, 200).getRGB();
+			//int diffColor = new Color(255, 0, 0, 200).getRGB();
```
bruderol: no dead comments please

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
```diff
@@ -88,7 +85,8 @@ public double compare(final String baseUseCaseName, final String baseScenarioNam
...
+			int diffColor = parameter.getDiffImageColor().getRGB();
```
bruderol: funny thing here --- so we are transforming a string to a BigInteger and then into an AWT color - just to in the end get out an int of it again? Couldnt we just use string for the color everywhere, and only here in this place we transform it to an Integer by using Integer.decode ??

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/AbstractComparator.java
kabl: The main problem is that before were three parameters. And this three parameters are passed through five classes at least. After adding one more value had to modify 5 classes and several test classes. Now pass through a configuration class. So no need to change so many files if adding or removing a parameter.

scenarioo-server/src/main/java/org/scenarioo/model/configuration/Configuration.java
kabl: That's kind of the default import strategy from IntelliJ. Fixed

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ComparatorParameter.java
kabl: Fixed

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
kabl: removed

scenarioo-server/src/main/java/org/scenarioo/model/configuration/Configuration.java
kabl: Changed to "Long.decode(diffImageColorRgbaHex).intValue();"

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
kabl: That would be possible. IMHO the configuration should provide a Color (as it is now). It's the Clients decision what he does with the Color. Otherwise the Client needs knowledge about configuration and that it's some magic Hex String...

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ComparatorParameter.java
kabl: OK

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/comparator/ScreenshotComparator.java
kabl: ok

PR #611, opened 2017-05-09
Declare security role in web.xml

samcontesse:
Avoid warning during Tomcat startup.

From the web-app.xsd documentation :

The role-name used here must either correspond to the role-name of one
of the security-role elements defined for this web application,
or be the specially	reserved role-name "*" that is a compact syntax for
indicating all roles in the web application.


bruderol:
@samcontesse I added a description to the security role in the web.xml and will merge that now. Thanks a lot!

PR #632, opened 2017-10-08
Feature/#603 improve comparison view

bruderol:
Improvements of Step Comparison View
* better buttons to switch comparison views
* legend for changes highlights in screens
* adjusted default color for change highlights in screens
* simplified code

!! only to be merges after other open PR by @mi-we and @sascha-kaufmann is merged !!


bruderol:
wrong branch again 🙈

Issue #603, opened 2017-03-27
Adjust Step Comparison View


bruderol:
The Step Comparison View needs to be reviewed and slightly improved after latest changes to the comparisons.

Acceptance criterias
* order and logic of behaviour of buttons makes sense and is understandable to the user (to be reviewed by Rolf and some other users)
     * also check that it is understandable, in case there are no changes at all or one of the screenshots is just missing (added or removed step)
* the labels above the screenshots needs to be improved: currently it is not easily understandable in all cases which screenshot shows which version and it is also not very consistent what is displayed
    * the label displayed should correspond to the text that is displayed in the dropdown selection more or less
    * displaying a branch or build alias might not make much sense here
    * keyword "Current" might make sense to additionaly mark which screenshot is the current selected build
    * it should not depend on what is displayed whether the comparison build was configured using an alias, in any case the concrete used build/branch label should be visible
* the URL of the view should as well contain the selected tab, such that a link to the comparison view can be sent by link.

This story might need clarification with @bruderol on some detail design by reviewing the current situation.

adiherzog:
With the diff viewer feature there are now many layers of menus, tabs and buttons on the step view. I started making some improvements:

* Breadcrumbs are now right below the navigation bar. This makes the screen look a bit calmer and saves some vertical pixels.
* Rephrased "Substep 1 of 3 on current page" to "Step 1 of 3 on Page" (shorter) and included it in the step navigation. As this text sometimes led to a line break, this saves some space. It also simplifies the UI.
* Moved the "Show Highlights" button all the way to the right, so that it does not jump around when switching to the single page view.
* Removed the text "highlighted changes", as this should already be clear by the state of the "Hide Highlights" button.

This is not checked in yet. @bruderol @danielsuter How do you like it? Any more ideas how we can simplify the UI, gain space or make it better understandable?

At the moment:
![image](https://user-images.githubusercontent.com/3780183/28932012-59fc6f92-7878-11e7-9764-88b7ebd96f2b.png)

New:

![image](https://user-images.githubusercontent.com/3780183/28932039-6fa38092-7878-11e7-967c-b34eefeab081.png)

adiherzog:
I commited it. We have to further improve the main navigation bar, because with the comparison dropdown and the search field shown, it results in two rows, which looks quite ugly.

bruderol:
[...]

PR #639, opened 2017-10-23
Feature/603 improve comparison view

bruderol:
#603 Improvements of step comparison view:
* better buttons to switch comparison views
* display legend for changes highlights in screens
* display as well a special legend for screens without changes or added steps
* adjusted default color for change highlights in screens (use same as for marking changes in all other pages, adjusted this other color as well to be slightly darker/stronger)
* simplified code of steps page (use explicit controller functions instead of complex generic code)

finally a PR with only my changes and conflicts resolved - ready to review and merge!

You can review the changed view with new button in this example here:
http://demo.scenarioo.org/scenarioo-feature-603-improve-comparison-view/#/step/Donate/find_donate_page/searchResults.jsp/0/0?branch=Production&build=last%20successful&comparison=To%20Projectstart

### DIFF
--- a/scenarioo-client/app/step/step.controller.js
+++ b/scenarioo-client/app/step/step.controller.js
@@ -21,7 +21,7 @@ function StepController($scope, $routeParams, $location, $route, StepResource, H
                         $filter, ApplicationInfoPopupService, GlobalHotkeysService, LabelConfigurationsResource,
                         SharePageService, SketcherContextService, RelatedIssueResource, SketchIdsResource,
                         SketcherLinkService, BranchesAndBuildsService, ScreenshotUrlService, SelectedComparison, BuildDiffInfoResource,
-                        StepDiffInfoResource, DiffInfoService, localStorageService) {
+                        StepDiffInfoResource, DiffInfoService, localStorageService, ConfigService) {
 
     var transformMetadataToTreeArray = $filter('scMetadataTreeListCreator');
     var transformMetadataToTree = $filter('scMetadataTreeCreator');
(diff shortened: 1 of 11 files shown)


bruderol:
@adiherzog and/or @danielsuter please review and merge!

scenarioo-client/app/step/step.controller.js
```diff
@@ -21,7 +21,7 @@ function StepController($scope, $routeParams, $location, $route, StepResource, H
                         $filter, ApplicationInfoPopupService, GlobalHotkeysService, LabelConfigurationsResource,
                         SharePageService, SketcherContextService, RelatedIssueResource, SketchIdsResource,
                         SketcherLinkService, BranchesAndBuildsService, ScreenshotUrlService, SelectedComparison, BuildDiffInfoResource,
-                        StepDiffInfoResource, DiffInfoService, localStorageService) {
+                        StepDiffInfoResource, DiffInfoService, localStorageService, ConfigService) {
```
danielsuter: Yet another dependency. Couldn't it be integrated into an existing service?

scenarioo-client/app/step/step.controller.js
```diff
@@ -464,35 +457,50 @@ function StepController($scope, $routeParams, $location, $route, StepResource, H
...
+    $scope.toggleComparisonChangesHighlighted = function() {
+        $scope.comparisonViewOptions.changesHighlighted = !$scope.comparisonViewOptions.changesHighlighted;
+        setLocalStorageValue('diffViewerStepComparisonChangesHighlighted', $scope.comparisonViewOptions.changesHighlighted);
+    };
+
+    $scope.getComparisonViewHighlightChangesColor = function() {
+        return ConfigService.diffViewerDiffImageColor();
+    };
+
+    function getLocalStorageBool(storageKey){
```
danielsuter: Can't we move it into the local storage service? So we could also unit test that one.

scenarioo-client/app/styles/scenarioo.less
```diff
@@ -78,6 +78,14 @@ h4 {
   width: 150px !important;
 }
 
+.btn-default.focus, .btn-default:focus {
```
danielsuter: Is a global override wanted here?

bruderol: yes, I think it makes sense. Having same background color for focused buttons as if they were toggled makes no sense to me. Still the focused buttons are identifieable, because it gets a different border color

scenarioo-client/app/step/step.controller.js
bruderol: @danielsuter haha, are you joking?? ;-) there are tooooo much dependencies already, sure. but this component needs more serious refactoring than what I had time for. I think the whole step component should be broken up into sub components. E.g. the comparison tab should be its own component (In my oppinion: should have been designed like that right from the start!). And we should also think about refactoring some of the dependencies, but I definitvely didnt have time for refactoring all that as part of the small changes I wanted to do. If you see how this dependency could have been avoided very easily by not introducing too much needless complexity, feel free to change it! To me it doesnt look reasonable to get the property from one of the other services. None of them sounds the place to be for that. And also I do not like to touch two kind of services just for passing a simple config property into the controller or view where it is needed. So all in all: good topic to discuss once, but not something to be changed fast without discussing together and thinking about what would be the best structure for the future. Maybe you want to open an issue to refactor that once?

scenarioo-client/app/step/step.controller.js
bruderol: yes, you are right on that - feel free to change - I have no time for it. I think some of that code I refactored here -
 especially concerning that access of properies from local storage - could have been done much better in the original code. Used me a lot of time because that was not designed in a way that was easy to change, I think I improved it already a lot. We can improve it even more of course!

bruderol:
@danielsuter I think none of your review comments - although valid comments - should be blocking for meging the PR. See my comments. So could you please click the merge button? We can still improve on the points you mentioned (if anybody has time) later. Thanks.

bruderol:
@danielsuter whahaaaa that was fast reaction! Impressive! Greetings to Belgrade! 😄

bruderol:
@danielsuter Thanks a lot!

Issue #630, opened 2017-10-05
Hide Screen Annotation Button without Annotations


bruderol:
The button is displayed wven if there are no annotations at all.

This was once different!

Important to not display it, if there are no annotations.

bruderol:
fixed by @ideco - Thanks @ideco - please see my comments on PR #647 
I also added the fix to the changelog.

PR #647, opened 2017-11-12
Fix the visibility of the "Show Screen Annotations"-Button

ideco:
The button is now only displayed when the screenAnnotations array
actually contains elements and not if it is just present.


ideco:
Turns out this was easier than expected

scenarioo-client/app/step/screenAnnotations/screenAnnotationsButton.html
```diff
@@ -1,7 +1,7 @@
 <!-- View Screen Annotations toggle button -->
-<div style="display:inline-block;" id="sc-showHideScreenAnnotationsButton" ng-click="screenAnnotationsButton.toggleAnnotationsVisible()" ng-show="screenAnnotations">
+<div style="display:inline-block;" id="sc-showHideScreenAnnotationsButton" ng-click="screenAnnotationsButton.toggleAnnotationsVisible()" ng-show="screenAnnotations.length">
```
bruderol: I think it would be better to implement a 'hasScreenAnnotations' function on the directive that encapsulates an expression that also checks screenAnnotations to not being null or undefined - that would be more robust - otherwise this code relies on the list always to be an array (did you check that this is really the case?)

bruderol:
@ideco Let me know when you have the test done - looks good so far! And please also push the other small code improvements we did in the review of existing controller code. Thanks!

bruderol:
@ideco seems there are some tests failing - your bug fix was not yet correct, sorry - I found what was missing - will fix it for you and add some explanation to the fix later. Is some JS glitch.

scenarioo-client/app/step/screenAnnotations/screenAnnotationsButton.html
```diff
@@ -1,5 +1,5 @@
 <!-- View Screen Annotations toggle button -->
-<div style="display:inline-block;" id="sc-showHideScreenAnnotationsButton" ng-click="screenAnnotationsButton.toggleAnnotationsVisible()" ng-show="screenAnnotationsButton.isAnnotationsButtonVisible">
+<div style="display:inline-block;" id="sc-showHideScreenAnnotationsButton" ng-click="screenAnnotationsButton.toggleAnnotationsVisible()" ng-show="screenAnnotationsButton.isAnnotationsButtonVisible()">
```
bruderol: there was a `()` missing - without `()` it will only return the function without calling the function - which in this case is again an expression that is almist always true as long as this function is defined.

bruderol:
this fixes #630 - I will merge it - please @ideco see my important comments and what I had to fix! Thanks! 

Thanks for your contribution! Looking forward to more of your great contributions in 2018! :-)

PR #661, opened 2018-01-06
#660 Improve step comparison screenshot captions

### LINKED ISSUE #660: Improve Step Comparison View Screenshot Captions
As user I do not want the screen to get cluttered with the full timestamp and revision above each of the screenshots in comparison view, such that I see more clearly the most important information (which screenshot is which build) and the layout does not get broken even on smaller screens too fast.

Acceptance criteria
* each screenshot caption starts with short prefix:
    * "Current:" for current screenshot of current selected build
    * name of the selected comparison for the screenshot of the other build that it is compared to
* behind this prefix only the name of the build (and if available the alias of the build) is displayed above each of the screenshots followed by an information icon, to make it obvious that there is more information about the build that can be showed in tooltip over that icon.
* Less important information about each screenshot like timestamp  and revision is only displayed in tooltip over the information icon

bruderol:
with info icon and tooltip - fixes #660

also added e2e tets for comparison screenshot titles.

also fixes some issues with e2e tests that rely on not clearing local storage for each test, which should not be the case, no test should rely on anything done by another previous test, because that will lead to follow up issues once a test fails or modifies local storage to an unexpected state for next test (e.g. if it was on comparison tab and next test does not expect to have the comparison tab open).

also restructured the step tests slighlty to have separate test files (and thus "use cases") for comparison and normal step view, and improved naming to have all step tests start with prefix "step` (I think we should rename all test use cases to folow this rule, to start with the context in the name).


bruderol:
will merge it now - @adiherzog or all @scenarioo/scenarioo-devs you can review and give feedback if you like.

bruderol:
also added a test now for the screen titles.

then there was some trouble with the CI build - see here:
http://build.scenarioo.org/jenkins/job/scenarioo/job/feature%252F660-improve-screenshot-comparison-captions/

but on a copy of the same branch with the same commits it was able to build successfully:
http://build.scenarioo.org/jenkins/job/scenarioo/job/feature%252F660-improve-screenshot-comparison-captions-fix-build/

VERY STRANGE!

will merge it now - feedback still appreciated.

bruderol:
ok, now there is a problem with the new popup test - will fix it

bruderol:
@adiherzog and @danielsuter I fixed some issues here with e2e tests breaking once another test fails or does not leave in expected state - e.g. if comparison tab was selected, which is remembered in local storage. Now all tests rely on cleared local storage before starting, and all tests should be designed like that - no test should rely on any state set by other previous test.

PR #679, opened 2018-05-06
Improve E2E test documentation

adiherzog:


docs/contribute/e2eTesting.md
```diff
@@ -1,66 +1,69 @@
...
+
+
+### Manually update ChromeDriver
+
+If there's a problem with ChromeDriver, try the following commands.
+ 
+On Linux:
+```
+./node_modules/protractor/bin/webdriver-manager update --chrome
+```
```
bruderol: If we have that problem often we should add that command to the npm scripts - such that you can use something like `npm run webdriver-manager update --chrome` or similar - somebody already added a webdriver-manager alias in package.json - not sure whether it works like that - but it should kind of.

Since we are using the chrome also for windows I wonder whether it wouldnt work as well for windows, so we could simply change the "pretest-e2e" script in package.json accordingly.

docs/contribute/e2eTesting.md
```diff
@@ -1,66 +1,69 @@
...
+
+On Windows:
+```  
+node node_modules\protractor\node_modules\webdriver-manager update
+```
```
bruderol: There is a script "pretest-e2e" for this purpose, so this should not be needed.

bruderol:
@adiherzog please have a quick look at my late comments here. thanks.

PR #683, opened 2018-05-08
Feature/650 diff manage backend

bruderol:
backend with incomplete new view - notz ready to merge yet but maybe could be started to review the backend part - which should be okay


bruderol:
closing in favour of other inclusing final view

PR #696, opened 2018-05-12
Feature/604 background execution of comparisons

### LINKED ISSUE #604: DiffViewer Plus: Background Execution for Comparisons
Currently the scheduling and management of diff comparison jobs is suboptimal in several ways, that needs to be analysed and if possible improved.

**Acceptance Criteria**
* [x] A running build comparison should not block another pending build import to be calculated, both should be able to run in parallel (two independent background thread workers).
* [x] Only one automatic comparison should be calculated at once (as was before! - later we could add a config property to parallelize multiple image comparisons in one such comparison job, but this is out of scope here)
* [x] Be aware that the REST service is another place where comparisons can be triggered by build jobs just on demand, and those will run in parallel with everything else in any case (fixed / imprioved by @adiherzog already on #605)
* [x] Management view for running build comparisons: similar to what we have for imported builds, including the log that is already stored (done by @sascha-kaufmann, @mi-we, @adiherzog and @bruderol on #650).
* [x] Reimporting a build only triggers new comparisons but not recalculation of allready existing comparisons (only those that do not have a build.xml in comparison folder yet, no matter what is the status!). (see also #624)
* [x] When a calculation background job starts, it only does something if the calculation is still in state QUEUED_FOR_PROCESSING - otherwise it means that it was already calculated (probably submitted for execution more than once).  
* Out of Scope: Avoid potential concurrency problem: how to cancel a running comparison job, in case somebody triggered to reimport one of the builds to compare? 
    * currently we ignore this issue - worst thing that can happen is that a wrong comparison is calculated or that it fails (which the user would see in new management view and can retrigger calculation) - but I expect that in most cases it could still calculate the comparison and even store it without problems.
* Out of Scope: this story does not address the background jobs of search indexing
* [x] adjust the related story #566 after this story here is completed, to only cover the remaining issue of search index calculation (currently done inside the import itself).

Thanks to work by  @mi-we, @sascha-kaufmann, @adiherzog and @bruderol for issues #650 and #605 this story is now easy to resolve and I will put comparison calculation into a separate thread executor now, to have that properly paralellized with build imports.

bruderol:
Fixes #604 

Separate execution queue for parallel comparison of builds while others are imported.


bruderol:
I will open a new one for all remaining Diff Viewer Plus Changes

Issue #686, opened 2018-05-08
Update nodejs to 8.x LTS version


adiherzog:
Reason: currently the package-lock.json is not generated and all our dependencies can change every time we run a build. This is bad because it means no reliable reproducible builds.

Version of nodejs we want to upgrade to: **8.11.2** (LTS version as of 21.5.2018).

# Tasks

* [x] Update in code / locally
* [x] Run E2E tests locally
* [x] Test it locally
* [x] Create backup of CI image before update
* [x] Update on CI
* [x] Make node dependency everywhere a fix dependency to the exact same specific version.
* [x] Document and inform the team by mail that they have to use that specific version as much as possible and how they can change the version if needed (locally, on CI, in repo, ...).

Drawback will be that debugging of E2E tests will not work like now anymore.
We have to handle that later see ssue #676.

bruderol:
Drawback of not having the package lock on the other hand is that our build is much less reliable - cause anytime it could use different transitive dependencies

I prefer get rid of that drawback and have a little drawback with debugging e2e tests - there is a potential workaround for (as discussed).

adiherzog:
Started implementing and testing this on branch `feature/686-nodejs-8`. The CI build for the branch fails:

http://build.scenarioo.org/jenkins/job/scenarioo/job/feature%252F686-nodejs-8/

Next step will be to update node on CI.

adiherzog:
Wrote e-mail to the google group. Will work on https://github.com/scenarioo/scenarioo/issues/676 next and inform regarding debugging as part of this issue.

PR #706, opened 2018-05-28
#686 Upgrade to nodejs 8.11.2

adiherzog:


bruderol:
Please document how we can debug e2e tests currently without ocntrol flow debugging using my special trick - thanks!

PR #709, opened 2018-05-30
WIP Feature/676 disable promise manager

adiherzog:


scenarioo-client/test/protractorE2E/specs/list_scenarios.ts
```diff
@@ -1,79 +1,78 @@
...
-                step('all pages are collapsed by default, "expand all" button is visible');
-                scenarioPage.expectOnlyExpandAllButtonIsDisplayed();
-                scenarioPage.toggleShowAllStepsOfPage(0);
-                step('"expand all" button and "collapse all" button are both visible');
-                scenarioPage.expectExpandAllAndCollapseAllButtonBothDisplayed();
-                scenarioPage.toggleShowAllStepsOfPage(1);
-                step('Only "collapse all" visible');
-                scenarioPage.expectOnlyCollapseAllButtonIsDisplayed();
+            .it(async () => {
+                await Utils.navigateToRoute('/');
```
bruderol: I think `HomePage.goToPage()` was nicer than a direct navigateToRoute

bruderol: this has been changed in many other places.

scenarioo-client/test/protractorE2E/util/util.ts
```diff
@@ -0,0 +1,112 @@
...
+export async function refreshBrowser() {
+    return browser.navigate().refresh();
+}
+
+export async function assertElementIsEnabled(elementId) {
+    return expect(element(by.id(elementId)).getAttribute('disabled')).toBeFalsy();
+}
+
+export async function assertElementIsDisabled(elementId) {
+    return expect(element(by.id(elementId)).getAttribute('disabled')).toBeTruthy();
```
bruderol: I know you did not write this - but a remark for future improvements - I think it would be better style to not pass ids or css classes to helper methods to identify elements, but use elements directly as params and pass them to such similar helper methods or even components. 

Advantage: these same functions can then be used no matter whether I used an id or a css selector or whatsoever to resolve the element. Also elements can even be easily reused in protractor, no problem at all.  

In general I think WebdriverJS elements are already a nice way of defining elements and we do not need to replace / wrap that by our own methods.

scenarioo-client/test/protractorE2E/webPages/branchAliasesPage.ts
```diff
@@ -1,113 +1,93 @@
...
-    } else {
-        BaseWebPage.call(this, '/manage?tab=branchAliases');
-    }
-
-    this.branchAliasTable = element(by.css('table.table-responsive'));
-    this.saveButton = element(by.css('input.btn[value="Save"]'));
-    this.resetButton = element(by.css('input.btn[value="Reset"]'));
-
-    this.branchesDropDown = element(by.css('div.navbar ul.nav li.dropdown ul.dropdown-menu'));
+    private branchAliasTable: ElementFinder = $('table.table-responsive');
```
bruderol: thanks to type inference I guess that the type `ElementFinder` should be somehow needless here - I think we could ommit that on all such private fields, as long as we initialize them with a concrete object of correct type already (as here and usually for all such fields).

scenarioo-client/test/protractorE2E/specs/list_scenarios.ts
adiherzog: Agree, done.

scenarioo-client/test/protractorE2E/util/util.ts
adiherzog: Agree, but won't change as part of this PR as it's not in the scope of this change.

scenarioo-client/test/protractorE2E/webPages/branchAliasesPage.ts
adiherzog: Agree, done.

adiherzog:
DO NOT DELETE THIS BRANCH, IT'S GOOD FOR DEMO PURPOSES OF THE DIFF VIEWER.

scenarioo-client/test/protractorE2E/util/util.ts
bruderol: sure

bruderol:
@adiherzog you can secure branches from being deleted under settings!

adiherzog:
@bruderol Thanks for the tip, did not know that.

bruderol:
@adiherzog shouldnt we also merge this to the release branch ?? or was it something risky not to put into the release ??

bruderol:
@adiherzog removed my comment - this was already merged before - all ok - sorry.

PR #715, opened 2018-06-01
#673 Configure Comparisons with regexp for all feature branches

### LINKED ISSUE #673: Comparison configurations with regexp for base branch
As a user I want to be able to define comparisons for all feature branches against the develop branch.

# Acceptance Criteria

* Allow regular expressions for the base branch of a comparison configuration
* Pure branch names as base branch definitions like "feature/some-feature" should work as before

bruderol:
Fixes #673 

Now we can use regexp in comparison configurations like "feature/.*" to compare all feature branches against e.g. "develop".

I configured for our self docu to always compare feature branches against develop on every new build.

### DIFF
--- a/scenarioo-docu-generation-example/src/test/resources/config-for-demo/config.xml
+++ b/scenarioo-docu-generation-example/src/test/resources/config-for-demo/config.xml
@@ -59,7 +59,7 @@
 	<comparisonConfigurations>
         <comparisonConfiguration>
             <name>To Develop</name>
-            <baseBranchName>scenarioo-feature-*</baseBranchName>
+            <baseBranchName>scenarioo-feature-.*</baseBranchName>
             <comparisonBranchName>scenarioo-develop</comparisonBranchName>
             <comparisonBuildName>last successful</comparisonBuildName>
         </comparisonConfiguration>
--- a/scenarioo-docu-generation-example/src/test/resources/config-for-demo/config.xml
+++ b/scenarioo-docu-generation-example/src/test/resources/config-for-demo/config.xml
@@ -58,6 +58,18 @@
     </branchAliases>
 	<comparisonConfigurations>
         <comparisonConfiguration>
+            <name>To Develop</name>
+            <baseBranchName>scenarioo-feature-*</baseBranchName>
+            <comparisonBranchName>scenarioo-develop</comparisonBranchName>
+            <comparisonBuildName>last successful</comparisonBuildName>
+        </comparisonConfiguration>
+        <comparisonConfiguration>
+            <name>To Last Release</name>
+            <baseBranchName>scenarioo-develop</baseBranchName>
+            <comparisonBranchName>scenarioo-master</comparisonBranchName>
+            <comparisonBuildName>last successful</comparisonBuildName>
+        </comparisonConfiguration>
+        <comparisonConfiguration>
             <name>To last successful</name>
             <baseBranchName>wikipedia-docu-example-dev</baseBranchName>
             <comparisonBranchName>wikipedia-docu-example</comparisonBranchName>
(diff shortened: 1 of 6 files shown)


scenarioo-docu-generation-example/src/test/resources/config-for-demo/config.xml
```diff
@@ -57,6 +57,18 @@
         </branchAlias>
     </branchAliases>
 	<comparisonConfigurations>
+        <comparisonConfiguration>
+            <name>To Develop</name>
+            <baseBranchName>scenarioo-feature-*</baseBranchName>
```
adiherzog: Shouldn't this be `.*` instead of just `*`?

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/ComparisonExecutor.java
```diff
@@ -336,7 +336,8 @@ private void logDuration(long startTime) {
 		String resolvedBaseBranchName = aliasResolver.resolveBranchAlias(baseBranchName);
 		for (ComparisonConfiguration comparisonConfiguration : comparisonConfigurations) {
 			String resolvedComparisonBranchName = aliasResolver.resolveBranchAlias(comparisonConfiguration.getBaseBranchName());
-			if (resolvedBaseBranchName.equals(resolvedComparisonBranchName)) {
+			if (resolvedBaseBranchName.equals(resolvedComparisonBranchName)
+				|| resolvedBaseBranchName.matches(resolvedComparisonBranchName)) {
```
adiherzog: Please write a unit test for this. That will probably reveal a bug.

I think what you wanted to write was `resolvedComparisonBranchName.matches(resolvedBaseBranchName)`.

https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#matches(java.lang.String)

scenarioo-docu-generation-example/src/test/resources/config-for-demo/config.xml
bruderol: Yes, should! Thought I did that -will check ...

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/ComparisonExecutor.java
bruderol: @adiherzog the `reseolvedComparisonBranchName` is the one from the configuration, which is the regexp, which needs to go into the parameter, so I think it is correct and it worked when I tested it.

bruderol:
@adiherzog I corrected the config, but the impl is okay - you can verify it on the demo as follows - if you see comparisons "To Projectstart" here:
http://demo.scenarioo.org/scenarioo-feature-673-configure-automatic-comparisons-for-all-feature-branches-with-regexp/#/manage?branch=wikipedia-docu-example&build=last%20successful&comparison=Disabled&tab=comparisons

These have been configured with a regexp on all wikipedia branches, as follows:
```
<comparisonConfiguration>
            <name>To Projectstart</name>
            <baseBranchName>wikipedia-docu-.*</baseBranchName>
            <comparisonBranchName>Production</comparisonBranchName>
            <comparisonBuildName>2014-01-20</comparisonBuildName>
        </comparisonConfiguration>
```

And if that hadnt work, there would for sure some e2e tests fail that expect those comparisons to be present.

I will see if I can easily add a unit test, but if not, I think that should be more than ok to merge.

adiherzog:
@bruderol Now I get it. What confused me was that `resolvedComparisonBranchName` does not sound like "base branch". Could you please rename this to `resolvedComparisonBaseBranchName`. Thanks.

bruderol:
@adiherzog Yeah I see your point of confusion - I also had same troubles - will try to improve it accordingly. Also I improved my impl slightly and unit tested it now a little bit. Its not perfect but at least it is covered.

But really I have to stop here now, this caused me more time than I wanted to spend on that.

bruderol:
@adiherzog Ok, I improved it - will merge it once the build is green - do you want to review what I improved?

scenarioo-server/src/main/java/org/scenarioo/business/diffViewer/ComparisonExecutor.java
```diff
@@ -327,17 +321,17 @@ private void logDuration(long startTime) {
...
 		List<ComparisonConfiguration> comparisonConfigurations = configurationRepository.getConfiguration()
 			.getComparisonConfigurations();
 		String resolvedBaseBranchName = aliasResolver.resolveBranchAlias(baseBranchName);
 		for (ComparisonConfiguration comparisonConfiguration : comparisonConfigurations) {
-			String resolvedComparisonBranchName = aliasResolver.resolveBranchAlias(comparisonConfiguration.getBaseBranchName());
-			if (resolvedBaseBranchName.equals(resolvedComparisonBranchName)) {
-				comparisonConfigurationsForBaseBranch.add(comparisonConfiguration);
+			String resolvedConfigurationBaseBranchName = aliasResolver.resolveBranchAlias(comparisonConfiguration.getBaseBranchName());
+			if (resolvedBaseBranchName.equals(resolvedConfigurationBaseBranchName)
+				|| resolvedBaseBranchName.matches(resolvedConfigurationBaseBranchName)) {
```
adiherzog: Nice, thanks for renaming it!

PR #718, opened 2018-06-01
CI Pipeline: also lock tomcat for e2e tests to not execute tests in parallel

bruderol:
with other deployments or other e2e tests against same tomcat (in other parallel build jobs)

@adiherzog maybe for now better to reduce troubles when we work in parallel.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -98,7 +98,7 @@ timestamps {
             ansiColor('xterm') {
 
                 try {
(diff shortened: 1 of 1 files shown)


bruderol:
@adiherzog although I have the impression that it was not a parallelization problem what we have seen today, but more a memory leaking problem on tomcat - we have to keep an eye on it, next time it happens.

PR #721, opened 2018-06-02
#720 Selfdocu: Compare master and develop on same demo

### LINKED ISSUE #720: Scenarioo Selfdocu: Compare master and develop - deploy them to same demo
I noticed that master is not deployed in our develop demo. Therefore the comparison with master does not work. See failed imports in Comparison tab: http://demo.scenarioo.org/scenarioo-develop/#/manage?branch=scenarioo-develop&build=last%20successful&comparison=Disabled&tab=comparisons

@bruderol Will the master documentation be deployed to develop after the release? If not, should it be?

bruderol:
Fixes #720 

Deploy self docu of master and develop always to both: master and develop demo instances.

This can also help to see whether our new version is backward compatible for old docu data from master branch or whether we introduced backwards compatible problems on master demo.

One open point: 
Does it make sense to deploy both to both our master as well as our develop instance?
* Pro: 
     1. We can also demo in the master demo for latest release, how it looks if two real builds are compared. 
     2. People can see in our master demo the newest upcoming features of scenarioo
     3. This can be tested on master demo, because on develop the master docu will still be missing until we merge that change to master (or we have to fix it once manually by copying the master branch to the develop demo)

* Contra: The new docu from the develop instance might become breaking (once we do breaking changes to the scenarioo format, which happens rarely and only for special releases) and not work on the master instance, We might have to disable that in such cases later again. 

I went for the 3 pro arguments outweighing the disadvantage that it might lead to failing imports on the master demo (which by the way is not that bad - would not fail our tests, since those do not test our self docu deployments)

@adiherzog what do you think?

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -40,7 +40,7 @@ BRANCH_DATA_ARCHIVE_DIR=/var/lib/scenarioo/data-archive/$BRANCH
 echo "Undeploy Old Scenarioo Web App"
 curl -u $TOMCAT_USERPASS http://localhost:8080/manager/text/undeploy\?path\=/scenarioo-$BRANCH
 
-# Backup all self docu and other important example docu reports to be restored after succesful deployment and testing of this scenarioo instance
+# Backup all self docu and other important example docu reports to be restored after successful deployment and testing of this scenarioo instance
 mkdir -p $BRANCH_DATA_ARCHIVE_DIR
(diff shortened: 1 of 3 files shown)


adiherzog:
@bruderol Thanks for the PR. I think it's good to deploy the develop and master self-docu to both develop and master.

As the master build will now fail (missing Elasticsearch 2) we should copy one master self-docu build to develop manually, just to demo / test the Diff Viewer feature.

PR #735, opened 2018-06-21
Enable Chrome headless mode

### LINKED ISSUE #734: Run e2e tests in headless mode
Build successful: http://build.scenarioo.org/jenkins/job/scenarioo/job/feature%252Fheadless-chrome/

Performance is about the same as in headed mode, so we don't save time.

Advantage:

* When running the UI tests locally, you can keep working and do not get distracted by the tests. When running the tests in headed with Scenarioo enabled, the browser comes to the foreground every time a screenshot is taken. This does not happen in headless mode, as there is not visible browser.
* You can't accidentally break a test because you interacted with the browser while the test was running.
* We can remove xvfb from the CI. One thing less to maintain.

Disadvantage:

* When debugging locally, one has to disable headless mode in the protractor config file to make the browser visible again.

adiherzog:
See #734 for advantages / disadvantages I see. Should we merge this?
Fixes #734.


bruderol:
what if I want to debug as a developer and I want to see the browser --- please offer a different config for debugging without headless - I uise that often and I wouldnt want to change the config file all the time and then by accident commit it..

bruderol:
@adiherzog please see my comment

adiherzog:
@bruderol As just discussed I enabled headless mode only for e2e-scenarioo, not for e2e.

bruderol:
Alright, you could merge - we should maybe consider to remove Xvfb display in CI pipeline

PR #737, opened 2018-06-21
Feature/605 rest endpoint documentation

adiherzog:
I put the documentation in a separate file and linked it from the Diff Viewer page, as it is quite long.

### DIFF
--- a/docs/features/diff-viewer/diff-viewer-rest-api.md
+++ b/docs/features/diff-viewer/diff-viewer-rest-api.md
@@ -1,0 +1,96 @@
+# Diff Viewer REST API
+
+The Diff Viewer REST API is intended for people who want to include visual regression testing in their build pipeline. E.g. you could make your build fail if the overall change rate is beyond a certain threshold.
+
+Parameters:
+
+* `branchName` / `buildName` in URL: build to compare with the comparison build
+* `comparisonName`: name under which the comparison will be available for selection
+* `banchName` / `buildName` in payload: comparison build
+
+## Calculate Configured Comparisons
+
+The comparisons configured in the `config.xml` file are automatically calculated after a build import (see [Publish Scenarioo Documentation Data](../../tutorial/Publish-Documentation-Data.md) and [Configuration of Automatic Comparisons](diff-viewer.md#configuration-of-automatic-comparisons)). You can use a calculation endpoint (see below) to poll for completion of these calculations.
+
+## Import Build and Calculate Comparison Synchronously
+
+This call triggers a build import and comparison calculation and also blocks until both are done. This operation can take a long time for larger builds. Consider using the non-blocking API calls instead. 
+
+`POST /rest/builds/{branchName}/{buildName}/comparisons/{comparisonName}/importAndCompare`
+
+Payload:
+
+`{ "branchName": "{comparisonBranchName}", "buildName": "{comparisonBuildName}" }`
+
+In Scenarioo 3.0.0 the URL was different: `importBuild/{branchName}/{buildName}/{comparisonBranchName}/{comparisonBuildName}/{comparisonName}`.
+
+## Request Comparison Calculation
+
+Queues the calculation of a comparison between two builds. The method returns immediately and does not wait for the calculation to complete.
+
+`POST /rest/builds/{branchName}/{buildName}/comparisons/{comparisonName}/calculate`
+
+Payload:
+
+`{ "branchName": "{comparisonBranchName}", "buildName": "{comparisonBuildName}" }`
+
+
+## Get Calculation Status
+
+Returns the calculation status as a string. This is useful to check whether a queued calculation has already completed.
+
+`GET /rest/builds/{branchName}/{buildName}/comparisons/{comparisonName}/calculationStatus`
+
+The return value is one of these:
+
+* `QUEUED_FOR_PROCESSING`
+* `PROCESSING`
+* `SKIPPED`
+* `SUCCESS`
+* `FAILED`
+
+
+## Get Calculation
+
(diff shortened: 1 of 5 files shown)


docs/features/diff-viewer/diff-viewer.md
```diff
@@ -11,10 +11,11 @@ The following sections describe how you can create, browse, configure, manage an
...
 2. [Select Comparison to Browse](#select-comparison-to-browse)
 3. [Browse Comparison Differences](#browse-comparison-differences)
-4. [Hide Comparison](#hide-comparisons)
-3. [Configuration of Automatic Comparisons](#configuration-of-automatic-comparisons)
-4. [Manage Comparisons](#manage-comparisons)
-5. [Delete Comparisons](#delete-comparisons)
+4. [Hide Comparisons](#hide-comparisons)
+5. [Configuration of Automatic Comparisons](#configuration-of-automatic-comparisons)
+6. [Manage Comparisons](#manage-comparisons)
+7. [Integrate Comparisons in CI Pipeline](#integrate-comparisons-in-ci-pipeline)
```
bruderol: Please put this point below point 5 - since it is another way to trigger comparsions (similar to 5) and the manage view is less important.

docs/features/diff-viewer/diff-viewer.md
```diff
@@ -163,6 +164,10 @@ In the `Manage`-view there is a `Comparisons` tab that lists all calculated comp
 
 There you can also see error logs in case a calculation of a comparison failed in the `Details & Logs` dialog.
 
+## Integrate Comparisons in CI Pipeline
+
+See: [Diff Viewer REST API](diff-viewer-rest-api.md).
```
bruderol: A short explanation that there is a REST API which could be used in CI Pipelines to trigger comparisons (just 1 sentence) would be nice here.

docs/features/diff-viewer/diff-viewer-rest-api.md
```diff
@@ -0,0 +1,98 @@
...
+* `FAILED`
+
+
+## Get Calculation
+
+Returns the calculation object. This can be used to read the overall change rate (field `changeRate`).
+
+`GET /rest/builds/{branchName}/{buildName}/comparisons/{comparisonName}`
+
+Example response:
```
bruderol: I would remove that - one can try it out to get an example response and that docu would outdate probably fast - I want less overhead for maintaining our own docu. Thanks

docs/features/diff-viewer/diff-viewer-rest-api.md
```diff
@@ -0,0 +1,98 @@
...
+
+`GET /rest/builds/{branchName}/{buildName}/comparisons/{comparisonName}/calculationStatus`
+
+The return value is one of these:
+
+* `QUEUED_FOR_PROCESSING`
+* `PROCESSING`
+* `SKIPPED`
+* `SUCCESS`
+* `FAILED`
```
bruderol: Link to the enum on github for possible return values? Next time we change it, it will be wrong.

docs/features/diff-viewer/diff-viewer-rest-api.md
```diff
@@ -0,0 +1,98 @@
...
+
+`POST /rest/builds/{branchName}/{buildName}/comparisons/{comparisonName}/importAndCompare`
+
+Payload:
+
+`{ "branchName": "{comparisonBranchName}", "buildName": "{comparisonBuildName}" }`
+
+The response looks the same as the one returned by the [Get Calculation](#get-calculation) endpoint.
+
+In Scenarioo 3.0.0 the URL was different: `importBuild/{branchName}/{buildName}/{comparisonBranchName}/{comparisonBuildName}/{comparisonName}`.
```
bruderol: Please put that into the migration guide - in the docu we only document what is relevant for one version.

docs/features/diff-viewer/diff-viewer-rest-api.md
```diff
@@ -0,0 +1,98 @@
+# Diff Viewer REST API
+
+The Diff Viewer REST API is intended for people who want to include visual regression testing in their build pipeline. E.g. you could make your build fail if the overall change rate is beyond a certain threshold.
+
+Parameters:
+
+* `branchName` / `buildName` in URL: build to compare with the comparison build
```
bruderol: I would mention that this is the build where the comparison will be claculated on (where you can finally view it)

docs/features/diff-viewer/diff-viewer-rest-api.md
```diff
@@ -0,0 +1,98 @@
...
+* `comparisonName`: name under which the comparison will be available for selection
+* `banchName` / `buildName` in payload: comparison build
```
bruderol: something like "the build to compare with"

docs/features/diff-viewer/diff-viewer-rest-api.md
bruderol: I dont even think we had that at all in Scenarioo 3.0 since that was part of 4.0 to add this new service.

docs/features/diff-viewer/diff-viewer-rest-api.md
adiherzog: I think we should not change these, as it would be a breaking change. And to link to GitHub we would have to link to the correct branch. I'd rather leave it as is.

docs/features/diff-viewer/diff-viewer-rest-api.md
adiherzog: Ok, guess we never had this endpoint in an official version, so I'll remove it everywhere.

Issue #732, opened 2018-06-21
Success message after saving config not visible on small screen


adiherzog:
# Impact

Low, user could be unsure whether config was actually saved and might click "Save" again.

# Steps

* Go to "General Settings"
* Make config change
* Click "Save" -> Success message not visible because it is on the top of the page and not on the bottom where the Save button is.

bruderol:
I close this now.

PR #738, opened 2018-06-21
#732 Configuration Pages: Move success messages below save button

ideco:
This is a temporary fix for the problem, which simply moves success/failure messages below the save button in configuration page.

As a long term solution we should consider using toastr, once we've upgraded angular.


bruderol:
as discussed - you can merge afterwards :-) 👍

PR #741, opened 2018-06-21
Save disk space on CI server

adiherzog:


Jenkinsfile
```diff
@@ -59,8 +59,8 @@ properties([
 	buildDiscarder(logRotator(
 	    artifactDaysToKeepStr: '10',
 	    artifactNumToKeepStr: '5',
-	    daysToKeepStr: '30',
-	    numToKeepStr: '20'
```
bruderol: I still think that only 10 days is to short - sometimes there might be an issue that nobody has time to look at for 30 days - maybe even a not so easy to reproduce case  - but ok - we can increase it again when we need it

PR #748, opened 2018-06-22
Prevent Cleanup from removing important demos

bruderol:


bruderol:
@adiherzog I merge this now since I do not want to risk more cleanup of important demos

PR #760, opened 2018-10-12
Remove strange options from docker docu

bruderol:
I think we should not use docker options in our docu, that are not really needed to run our image - Users can decide on their own, what options they need if they want to tweak behaviour - the default is to keep the container.


bruderol:
In my oppinion for a real scenarioo installation you do not want the container to get removed after termination - why not keeping it once it was terminated - which by the way is the real default of docker (otherwise you wouldnt have to list it as an extra option) and I think it is a good default. People should decide on their own whether they want it to be removed automatically. If you look at other installation instructions of other services you wont find such options in their instructions as well - e.g. elasticsearch: https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html

PR #766, opened 2018-10-14
need to merge release branch to make demo on master work with elasticsearch changes

bruderol:
this is an exception - usually we wouldnt do that


bruderol:
will do it manually with fast forward

PR #784, opened 2018-12-05
Feature/778 spring endpoint security

### LINKED ISSUE #778: Spring Security for REST Endpoints
As application admin of Scenarioo I want to be still able to configure some users (username and password) also with new spring boot security mechanism in Scenarioo 5.0, such that the REST endpoints with write access can be secured.

**Acceptance Criteria**
* [x] One user and password can be configured to secure the build publishing endpoint
* [x] Requests to such secured REST endpoints are blocked, if no valid user is provided
* [x] Currently this is done only for the calls where we had a role defined before: probably only the build import REST endpoint
* [x] Per default these endpoints are secured with a random user name and password unless configured explicitly
* [x] How to configure fix usernames and passwords is documented well
  * [x] for spring boot setup
  * [x] for WAR deployment
  * [x] for docker setup
* [x] the breaking change is mentioned in changelog and shortly in the migration guide for version 5.0 (with link to the documentation)
* [x] it must be ensured that on the demo the same username and password is still configured for all demos (fix, not random, since used in other places, e.g. self-demo deployment)
* [x] tests that check
       * the call fails (e2e test) without correct user/pasword
       * the call succeeds with correct user/password

mi-we:
Fixes #778


bruderol:
@mi-we didnt review yet, but before we can merge this we have to address two things for sure:

* [ ] two unit tests seem to be very flaky on this branch and failed often on new CI build, we have to analyse this, it happens often.
* [ ] the target branch has not yet merged the develop branch, which is why some commits show up now (I, @bruderol , will take care of this second point, cause it needs manual merge!)

scenarioo-server/src/test/java/org/scenarioo/rest/builds/BuildsResourceIntegrationTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.io.IOException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@TestPropertySource(locations = "classpath:test-application.properties")
+public class BuildsResourceIntegrationTest {
```
bruderol: @mi-we The problem is that this test probably indirectly sets a context path on ContextPathHolder (a ugly singleton we should maybe get rid of) and once this test has been run, the StepIdentifierTest now fails because of different context path being initialized there.

bruderol:
@mi-we and @danielsuter I think we can merge for now, and address the documentation findings later.

PR #785, opened 2018-12-21
Enable OpenJDK 11 and Gradle 5

danielsuter:
Currently we have several problems:
- Gradle version is too low for OpenJDK 11 -> build fails
- Gradle files are incompatible with Gradle 5 -> build fails

This PR fixes all of this.

<< does not exist in Gradle 5 -> build not possible.
classesDir does not exist in Gradle 5 -> build not possible.
Gradle 4.7 is not compatible with Java 11 -> build not possible.
Wrapper task does not exist in Gradle 5 -> build not possible.


bruderol:
1. Are we sure that we want to update to JDK 11?? 
Will this mean that all projects need to run scenarioo on newest tomcat with java 11?
This would mean to adjust docu as well for developers as well as for users!
But first of all I would like to clarify whether it is okay to enforce Java 11 from our users???
@adiherzog what you think?


2. Why is the CI failing?

3. @danielsuter please merge develop again - you merged the other PRs in wrong order (the wrong one first) which is why one branch was not yet merged to develop and thus not to your branches as well. I corrected it on develop, please pull to your branches, such that you have all changes on these branches as well. Thanks.

build.gradle
```diff
@@ -51,8 +51,8 @@ allprojects {
 	group = 'org.scenarioo'
     version = getVersionName()
 
-    task wrapper(type: Wrapper) {
-        gradleVersion = '4.7'
+    wrapper {
+        gradleVersion = '4.10.3'
```
bruderol: I do not understand why this is still 4.10 although the PR talked about gradle 5

gradle.properties
```diff
@@ -1,3 +1,4 @@
 ossrhUsername=scenarioo
 # Set the password in your user home gradle.
-ossrhPassword=
\ No newline at end of file
+ossrhPassword=
+org.gradle.jvmargs=-Xmx4096m
```
bruderol: Do we really want to enforce this so high - will this work on CI??

See what we have in circleci config file

build.gradle
danielsuter: It's a deprecation. I did not upgrade to Gradle 5, simply made it compatible. I could not generate a wrapper anymore, because locally I'm running gradle 5. This does NOT use Gradle 5, just makes it Gradle 5 compatible.

danielsuter:
> 1. Are we sure that we want to update to JDK 11??
>    Will this mean that all projects need to run scenarioo on newest tomcat with java 11?
>    This would mean to adjust docu as well for developers as well as for users!
>    But first of all I would like to clarify whether it is okay to enforce Java 11 from our users???
>    @adiherzog what you think?
> 2. Why is the CI failing?
> 3. @danielsuter please merge develop again - you merged the other PRs in wrong order (the wrong one first) which is why one branch was not yet merged to develop and thus not to your branches as well. I corrected it on develop, please pull to your branches, such that you have all changes on these branches as well. Thanks.

This is not an upgrade to JDK 11, it simply makes it compatible. We should support Java 8 as long as possible as nobody is upgrading anyway. But I want to be able to compile scenarioo with a JDK 11 but Java 8 compatible locally.
Thanks I'll try to remerge develop.

danielsuter:
The only tests failing also fail on develop.

gradle.properties
```diff
@@ -1,3 +1,4 @@
 ossrhUsername=scenarioo
 # Set the password in your user home gradle.
-ossrhPassword=
\ No newline at end of file
+ossrhPassword=
+org.gradle.jvmargs=-Xmx600m
```
adiherzog: Why is the memory parameter required?

build.gradle
adiherzog: Gradle 5 supports Java 8, so we could actually update to Gradle 5. I would prefer if everybody would use the same gradle version, but it's up to you.

gradlew.bat
```diff
@@ -1,84 +1,84 @@
...
-rem Set variable GRADLE_EXIT_CONSOLE if you need the _script_ return code instead of
-rem the _cmd.exe /c_ return code!
-if  not "" == "%GRADLE_EXIT_CONSOLE%" exit 1
-exit /b 1
-
-:mainEnd
-if "%OS%"=="Windows_NT" endlocal
-
-:omega
+@if "%DEBUG%" == "" @echo off
```
adiherzog: Did the line endings change or why does the diff show all lines changed?

scenarioo-server/build.gradle
```diff
@@ -56,30 +56,19 @@ eclipse {
   }
 }
 
-task printPathConfigInformationForNodeJSAndGrunt << {
```
adiherzog: Guess this was an unused task

gradlew.bat
danielsuter: that's a generated file

gradle.properties
danielsuter: because we log a lot of stuff during the build which fills the memory

scenarioo-server/build.gradle
danielsuter: correct.

PR #788, opened 2018-12-30
Feature/787 test and fix self docu publishing

bruderol:
Work in progress ...

Fixes #788 

**Additionaly also adjusts the shared intelliJ settings (explicit project code style settings) and gitignore for these idea files** to only share what is explicitly meant to be shared. Removed some files not meant ot be shared (like gradle files and other files that depend on what other modules one has open in intellij, which is different for whatever one is working on)


bruderol:
@eggerschwiler thanks for fixing - seems there was a lot wrong - like @Body instead of @Param -I guess that was the key problem I missed/didnt see.

PR #790, opened 2019-02-01
Feature/789 build information in info popup

eggerschwiler:
The Info popup does not show the current version information because versions.properties was created too late.

Updated build.gradle that the necessary tasks are also created for bootWar- and not only for the war-task.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,10 +44,7 @@ jobs:
             export SCENARIOO_DATA=./scenarioo-docu-generation-example/build/scenarioDocuExample
             export SPRING_SECURITY_USER_PASSWORD=justfortest
             echo "Starting Scenarioo"
-            # To prevent a race condition between Spring Boot and ElasticSearch both trying to configure Netty
-            # we have to disable this through a System Property.
-            # See: https://discuss.elastic.co/t/contradictory-and-sometimes-poor-advice-given-for-es-set-netty-runtime-available-processors/148014
-            java -Xms800m -Xmx800m -Des.set.netty.runtime.available.processors=false -jar scenarioo-server/build/libs/scenarioo-latest.war > test-results/scenarioo-server.log || echo "Scenarioo server died. Exit code: $? see https://circleci.com/blog/how-to-handle-java-oom-errors/" >> test-results/scenarioo-server.log &
+            java -Xms800m -Xmx800m -jar scenarioo-server/build/libs/scenarioo-latest.war > test-results/scenarioo-server.log || echo "Scenarioo server died. Exit code: $? see https://circleci.com/blog/how-to-handle-java-oom-errors/" >> test-results/scenarioo-server.log &
 
             export PROTRACTOR_BASE_URL=http://localhost:8080/scenarioo
             export SCENARIOO_TARGET_DIRECTORY="../e2eScenariooDocu"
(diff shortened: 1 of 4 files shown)


eggerschwiler:
Additional changes in this branch:
- WebTest with DoubleClick fixed, it fails also on develop and on branches that were previously green
- System Property set when starting Spring Boot Fat War during CircleCI E2E-Tests to ensure that random Elasticsearch exception does not occur.

eggerschwiler:
Can be tested with this URL: http://demo.scenarioo.org/scenarioo-feature-789-build-information-in-info-popup/#/

.circleci/config.yml
```diff
@@ -44,7 +44,10 @@ jobs:
             export SCENARIOO_DATA=./scenarioo-docu-generation-example/build/scenarioDocuExample
             export SPRING_SECURITY_USER_PASSWORD=justfortest
             echo "Starting Scenarioo"
-            java -Xms800m -Xmx800m -jar scenarioo-server/build/libs/scenarioo-latest.war > test-results/scenarioo-server.log || echo "Scenarioo server died. Exit code: $? see https://circleci.com/blog/how-to-handle-java-oom-errors/" >> test-results/scenarioo-server.log &
+            # To prevent a race condition between Spring Boot and ElasticSearch both trying to configure Netty
+            # we have to disable this through a System Property.
+            # See: https://discuss.elastic.co/t/contradictory-and-sometimes-poor-advice-given-for-es-set-netty-runtime-available-processors/148014
+            java -Xms800m -Xmx800m -Des.set.netty.runtime.available.processors=false -jar scenarioo-server/build/libs/scenarioo-latest.war > test-results/scenarioo-server.log || echo "Scenarioo server died. Exit code: $? see https://circleci.com/blog/how-to-handle-java-oom-errors/" >> test-results/scenarioo-server.log &
```
bruderol: Couldnt we set this property somehow differently - this is very ugly if each user has to set this

scenarioo-client/test/e2e/pages/objectDetailsPage.ts
```diff
@@ -66,8 +66,10 @@ class ObjectDetailsPage {
 
         await Utils.waitForElementVisible(imageElement);
 
-        await imageElement.click();
-        return browser.actions().doubleClick(imageElement).perform();
+        // Browser action DoubleClick stopped working:
+        // return browser.actions().doubleClick(imageElement).perform();
+        // Workaround from https://stackoverflow.com/questions/44356348/selenium-3-x-double-click-not-working
+        return browser.executeScript("document.getElementById('" + imageId + "').dispatchEvent(new Event('dblclick'));");
```
bruderol: Thanks for the fix, but could we remove the unnecessary comments with dead code??

.circleci/config.yml
bruderol: So I would rather like to have a workaround in code that sets the property programatically.

eggerschwiler: I will have another look. The first couple of things I tried with a programatic solution didn't work, because the methods where I implemented a change were triggered too late.

scenarioo-client/test/e2e/pages/objectDetailsPage.ts
eggerschwiler: dead code removed and comments merged.

.circleci/config.yml
eggerschwiler: Solution with @PostConstruct in ScenariooViewerApplication worked. Reverted change in config.yml

bruderol:
Ich merge das mal - hab dafür nch nen neuen bug gefunden - siehe #792

PR #794, opened 2019-03-03
Feature/516 continue migration of files to Typescript and Angular 5

sebastianzillessen:
I continued the migration to Typescript and Angular 5.
- Splitting up restServices into one per file.
- Starting migration of services to Angular and making them available as downgraded components.

Please add me to the contributors list on scenarioo so that I don't need to use the fork method for new PRs. ☺️ 

Thanks


sebastianzillessen:
@danielsuter as we discussed last week. 
I started with some of the Rest Services. Work in progress!

PR #810, opened 2019-04-28
Feature/generate ts types

danielsuter:


scenarioo-server/build.gradle
```diff
@@ -24,6 +24,47 @@ apply plugin: 'eclipse-wtp'
...
+            'org.scenarioo.rest.objectRepository.StepReference',
+            'org.scenarioo.model.docu.aggregates.usecases.UseCaseScenarios',
+            'org.scenarioo.rest.scenario.dto.ScenarioDetails',
+            'org.scenarioo.rest.search.SearchResponse',
+            'org.scenarioo.rest.search.SearchEngineStatus',
+            'org.scenarioo.model.docu.aggregates.usecases.UseCaseSummary',
+    ]
+
+    addTypeNamePrefix = 'I'
+    noFileComment = true
```
sebastianzillessen: If you don't use it in an up to date check of gradle I would include the FileComment.

In the project where I use this concept we have deactivated the file comment because it prints a timestamp on every build, which triggers the up to date check of the gradle task and therefore increases the build time drastically.

scenarioo-server/build.gradle
```diff
@@ -24,6 +24,47 @@ apply plugin: 'eclipse-wtp'
...
+    outputFile = '../scenarioo-client/app/types/backend-types.ts'
```
sebastianzillessen: I would either put the output file in a folder "generated" or add generated to the name. 

And a good practice in TS is to name type definition files `*.d.ts` ☺️

scenarioo-server/build.gradle
```diff
@@ -24,6 +24,47 @@ apply plugin: 'eclipse-wtp'
...
+        jcenter()
+        mavenCentral()
+    }
+
+    dependencies {
+        classpath group: 'cz.habarta.typescript-generator', name: 'typescript-generator-gradle-plugin', version: '2.13.489'
+    }
+}
+generateTypeScript {
+    classes = [
```
sebastianzillessen: Cant you move all the backend DTOs into one package (`org.scenarioo.model`) and then import it from there? 

Listing every single file is dangerous imo as you expect new code to appear in your frontend Dtos as well.

scenarioo-server/build.gradle
```diff
@@ -24,6 +24,47 @@ apply plugin: 'eclipse-wtp'
 apply from: "${rootProject.projectDir}/build-java.gradle"
 apply plugin: 'org.springframework.boot'
 
+apply plugin: 'cz.habarta.typescript-generator'
+buildscript {
```
sebastianzillessen: Buildscript usually goes before applying any plugins as far as I know.

scenarioo-client/app/types/backend-types.ts
```diff
@@ -0,0 +1,278 @@
+/* tslint:disable */
+
+export interface IConfiguration {
```
sebastianzillessen: I think it would be great to use this type definitions already in the services. Otherwise they are not of much use... ☺️

scenarioo-server/build.gradle
danielsuter: I plan to do that in the future. Thus I already disabled it.

scenarioo-server/build.gradle
danielsuter: I will add the generated, good one! Do you have source for this good practise? Usually one uses d.ts files for libraries without types and normal types for your own code.

scenarioo-server/build.gradle
danielsuter: Unfortunately I would change the backend a lot. I don't want to do that. I'm aware that this is far from perfect but if somebody moves code, the build will break. So we should notice it.

scenarioo-server/build.gradle
danielsuter: Does not matter.

scenarioo-client/app/types/backend-types.ts
danielsuter: I'll do that in the other PRs. I want to have this change in develop first for everybody to use. So I can merge it into the other branches.

danielsuter: Done

PR #812, opened 2019-05-05
Feature/upgrade config service

danielsuter:
Not to be merged yet. Work in progress.


eggerschwiler:
@danielsuter: The changes look good to me. 
I looked at the test failures but only managed to fix one (applicationInfo.controller.spec.ts) and that one only used the ConfigMock and not the ConfigurationService.

danielsuter:
Will be merged after we've created a PR.

PR #824, opened 2019-05-14
Feature/717 diffviewer styling ui issues

### LINKED ISSUE #717: Diff Viewer: Styling and UI issues
Various layout issues and other ui problems on the step comparison view (diff viewer) and other screens with diff info indicators that are disturbing and lead to wrong interpretation - in prioritized order:

1. [important] when toggeling details to be displayed or hidden while on the comparison view the diff highlights are not properly aligned to the underlying screen (reload of page fixes it): 
image
   * [x] should be automatically alligned correctly without need to reload

3. [important] Change indicator tooltip hidden when details are not shown:
image
   * [x] user should see the tooltip on the change indicator on the screen (open to left)
 
4. [nice to have] Comparison view is difficult to understand for added screens - most buttons could be removed for added screens:
image
   * [x] Hide button for "Other Screen"
   --> see [comment](#issuecomment-492097689)
   * [x] Hide button to toggle between current and other screen
   * [x] Highlight the label "Added Step: No Comparison" by displaying it in green color (same as "added"-icon in tab)

5. [nice to have] Change indicator is currently empty for added screens on added scenario page:
image
   * [x] Display "added" for all steps in an added scenario (or leave the indicator away?? )

6. [nice to have] Use Case Table and Details overlap (probably because of the text length in the Description field). Same issue could be present in the scenarios list.
image
    * [x] Use Case and Scenario Table and Details on the right should never overlap, also not with diff info and long description texts in details

7. [nice to have] On the Page Details page, if any of Use Case, Scenario, Step or Page have a label, then all those headings appear, even for those that do not have labels.
177-labels-on-step-page
   * [x] Only the parts that have labels are listed.

8. [nice-to-have] Further issues without screenshots:
    * [x] After submitting a comparison using "Create Comparison", the Comparisons list should be displayed and refreshed to show that the comparison is being calculated (was not implemented on branch of this story, @bruderol will do it on other branch for #872)
    * [x] color of removed and added rows (use cases, scenarios) in tables is not very well distinguishable from normal row colors and especially not from current row selection when using up and down keys in tables - probably not at all distinguishable for color blind people. 
    * ~Add "Space" key to keyboard shortcuts documentation -> Toggles screenshots in step comparison view~ removed since not works without focus   --> comment by @eggerschwiler: "Space" only worked when I navigate with "Tab" to the button or when the button was pressed by a mouse click first. 

9. [nice-to-have] The "+" Tooltip is not completely visible on Scenario page.
717-show-hide-tooltip-not-shown
    * [x] Open tooltip to the right

Moved to other issues:

2. [important] Details move over the comparison screen on small screens ("responsive"):
image
   * [ ] details should move below the whole screen comparison area on small screen
   --> See [comment](#issuecomment-491750596)
   --> Moved to issue #826 because there is no quick and easy fix at the moment.

eggerschwiler:
fixes #717 

Postponed Issues:
* Issue 2: [important] Details move over the comparison screen on small screens ("responsive")
Reason: The problem is that if "Highlight Changes" is activated, then the screenshot-image has the following style: "position absolute". This does not work very well with responsive design.
The issue is even worse if the comparison type "Side-by-Side" is selected on a small screen. Then both screenshots are rendered in the same place and thus only the image normally on the right is visible.
--> Moved to separate issue #826 
* Issue 8.3: Add "Space" key to keyboard shortcuts documentation -> Toggles screenshots in step comparison view
Reason: Couldn't reproduce. "Space" only works when I navigate with "Tab" to the button or when the button was pressed by a mouse click first. In addittion, I didn't find the keyboard shortcuts documentation in Scenarioo Docs.
--> Shortcut not found in code, thus this will not be implemented.

Points to be discussed:
* Issue 4: "Comparison view is difficult to understand for added screens - most buttons could be removed for added screens."
Discussion: Removing the "Other Screen" button can lead to a confusing situation. For more details see [this comment](https://github.com/scenarioo/scenarioo/issues/717#issuecomment-492097689).
* Issue 8.2: "color of removed and added rows (use cases, scenarios) in tables is not very well distinguishable from normal row colors and especially not from current row selection when using up and down keys in tables - probably not at all distinguishable for color blind people."
Discussion: I changed the colors to be a bit brighter, to be more recognizable, however, it does not yet solve the problem for color blind people and I would like to get feedback on the colors used.
717-new-colors

All features can be reviewed in the [Demo](http://demo.scenarioo.org/scenarioo-feature-717-diffviewer-styling-ui-issues/?#/?branch=scenarioo-feature-717-diffviewer-styling-ui-issues&build=last%20successful&tab=usecases).
For issue 6, [build 768](http://demo.scenarioo.org/scenarioo-feature-717-diffviewer-styling-ui-issues/?#/?branch=scenarioo-feature-717-diffviewer-styling-ui-issues&build=build-768&tab=usecases&comparison=Disabled) has been modified to contain more labels.


scenarioo-client/app/manage/comparisons/createComparisonModal.controller.ts
```diff
@@ -200,6 +203,9 @@ function CreateComparisonModalController($uibModalInstance, BranchesAndBuildsSer
 
     function onSuccessCreation() {
         $uibModalInstance.close();
+        if (vm.onComparisonsPage) {
+            vm.route.reload();
```
danielsuter: will this do a browser refresh? Do we need that?

scenarioo-client/app/scenario/scenario.controller.ts
```diff
@@ -237,6 +257,17 @@ function ScenarioController($filter, $routeParams,
         return isUseCaseAdded;
     }
 
+    function isAddedScenario(useCaseDiffInfo) {
```
danielsuter: you may use ES6, we have babel to transpile. Alternatively you can migrate the file to typescript.

scenarioo-client/app/shared/metadata/metaDataButton.directive.ts
```diff
@@ -42,11 +42,17 @@ angular.module('scenarioo.directives').directive('scMetaDataButton', ($window, L
...
-        controller($scope) {
+        controller($scope, $route) {
             initMetadataVisibleFromLocalStorage($scope, $scope.localStorageKey);
             $scope.toggleShowingMetadata = () => {
                 $scope.linkingVariable = !$scope.linkingVariable;
                 LocalStorageService.set(STEP_METADATA_VISIBLE + $scope.localStorageKey, '' + $scope.linkingVariable);
+                // reload the page to ensure that highlighted changes are properly aligned
+                // if comparison highlights are activated and we are on the comparisons tab.
+                if ($route.current.scope.comparisonViewOptions && $route.current.scope.isComparisonChangesHighlighted()
+                    && sessionStorage.getItem('activeTab') === '2') {
```
danielsuter: Should a directive trigger a page reload?

scenarioo-client/test/e2e/pages/stepPage.ts
```diff
@@ -238,18 +238,22 @@ class StepPage {
...
-        await expect(element(by.id('sc-step-comparison-other-screen-view-button')).getAttribute('class')).not.toContain('active');
+        if (otherScreenVisible) {
+            await expect(element(by.id('sc-step-comparison-other-screen-view-button')).getAttribute('class')).not.toContain('active');
+        }
         await expect(element(by.id('sc-step-comparison-side-by-side-view')).isPresent()).toBeTruthy();
         return expect(element(by.id('sc-step-comparison-single-page-view')).isPresent()).toBeFalsy();
     }
 
-    async assertStepComparisonCurrentScreenViewIsActive() {
+    async assertStepComparisonCurrentScreenViewIsActive(otherScreenVisible: boolean) {
```
danielsuter: Imho this should be a separate method/function not a flag. The usage of the method is strange to me.

scenarioo-client/test/e2e/pages/stepPage.ts
```diff
@@ -238,18 +238,22 @@ class StepPage {
         return expect(element(by.id('comparison-tab')).isPresent()).toBeFalsy();
     }
 
-    async assertStepComparisonSideBySideViewIsActive() {
+    async assertStepComparisonSideBySideViewIsActive(otherScreenVisible: boolean) {
```
danielsuter: Imho this should be a separate method/function not a flag. The usage of the method is strange to me.

scenarioo-client/app/manage/comparisons/createComparisonModal.controller.ts
eggerschwiler: Yes. This will refresh the page. This is needed to solve issue 8.1 "After submitting a comparison using "Create Comparison", the Comparisons list should be refreshed (only when the user is on that tab)."

scenarioo-client/test/e2e/pages/stepPage.ts
eggerschwiler: Added separate function.

scenarioo-client/test/e2e/pages/stepPage.ts
eggerschwiler: Added separate function.

scenarioo-client/app/shared/metadata/metaDataButton.directive.ts
eggerschwiler: Good point! I moved the check and the reload to step controller.
However, we do need the page reload to solve issue 1: " when toggeling details to be displayed or hidden while on the comparison view the diff highlights are not properly aligned to the underlying screen (reload of page fixes it)"

scenarioo-client/app/scenario/scenario.controller.ts
eggerschwiler: Migrated the functions to ES6. I didn't want to upgrade to typescript, because I'm not 100% sure what else would need to be changed.

scenarioo-client/app/manage/comparisons/createComparisonModal.controller.ts
danielsuter: ok, I don't really understand why we can't just reload the data but need a page refresh since we are in a single page app.

eggerschwiler: I see your point. The difficulty I couldn't overcome was that we are here in the CreateComparisonModalController whereas the data would need to be reloaded in ComparisonsController. I couldn't find an example where we call one controller from another controller. Do you know how to do that?
As an aside, the refresh function in ComparisonsController also triggers a $route.reload(). The reason why it makes sense there (in my opinion) is that if we only load the data and the data hasn't changed, then there is no apparent change in the UI and the user might think that the refresh button didn't do anything and is broken, whereas if we reload the page then there is a visual feedback that something happened.

scenarioo-client/app/step/step.controller.ts
```diff
@@ -51,6 +52,16 @@ function StepController($scope, $routeParams, $location, $route, StepResource, S
         SelectedBranchAndBuildService.callOnSelectionChange(loadStep);
     }
 
+    /**
+     * reload the page to ensure that highlighted changes are properly aligned
+     * if comparison highlights are activated and we are on the comparisons tab.
+     */
+    function refreshIfComparisonActive() {
+        if ($scope.isComparisonChangesHighlighted() && $scope.activeTab === 2) {
+            $route.reload();
```
bruderol: I think it would be better to jump to the comparison page after this in any case and trigger a reload for sure, to give the user some feedback.

scenarioo-client/app/manage/comparisons/createComparisonModal.controller.ts
bruderol: I think we should instead just change the route to go to the comparison view in any case after creating a new comparison. will try to do it on a separate branch. is ok for now.,

scenarioo-client/app/manage/comparisons/createComparisonModal.controller.ts
```diff
@@ -39,6 +39,9 @@ function CreateComparisonModalController($uibModalInstance, BranchesAndBuildsSer
     vm.cancel = cancel;
     vm.createComparison = createComparison;
 
+    vm.onComparisonsPage = $routeParams['tab'] === 'comparisons';
+    vm.route = $route;
```
bruderol: I dont whink that injected services should be assigned to properties on the VM - those can be used directly from inside the controller can't they?

@danielsuter or what is your recommendation on this?

bruderol: will change this as well on my other branch

bruderol:
Thanks @eggerschwiler - Awesome work! :rocket:

PR #825, opened 2019-05-22
Feature/637 wait for startup calculations to complete

### LINKED ISSUE #637: Build Pipeline: Wait for all startup calculations completed and fail fast on error
Wait properly and stable for successful server startup before starting e2e tests, and avoid a fixed 2 minutes wait.

Acceptance criteria
* Ensure that tests will never start before server startup (including things like calculation of all diff comparisons) was done and succesful
* If a failure (exception log) occurs during waiting for the success of the startup then the build should immediately stop waiting any further and let the build fail 
* Wait for a well defined log entry that only occurs when all builds have been finished importing and all belonging diff comparisons have finished calculating (put an appropriate message for that into the application such that we can simply wait for this well defined message)

Purpose of thic change:
* currently there is a 2 minute sleep in the build script, that is not guaranteed to wait long enugh and still might often be longer than needed
* we want the build to fail as fast as possible if there occurs an exception during startup or import of builds
* waiting for one well defined startup and import success log entry should also make it more easier to change the things that are calculated during startup, and just waiting for that one well defined log entry (e.g. somewthing like "Scenarioo Startup COMPLETE - All import work COMPLETE") should be everything needed to wait for no matter how the app changes.

### LINKED ISSUE #704: Improve wait for server startup after deployment (e.g. wait for log statement)
Currently we just wait for a certain time (2 minutes?) which is not a very stable and makes builds unnecessarily long time blocking. 

Since It works pretty well at the moment though, so there's no high urgency to fix this.

eggerschwiler:
Wait properly and stable for successful server startup before starting e2e tests, and avoid a fixed 2 minutes wait.

Acceptance criteria
-  Ensure that tests will never start before server startup (including things like calculation of all diff comparisons) was done and succesful
-  If a failure (exception log) occurs during waiting for the success of the startup then the build should immediately stop waiting any further and let the build fail
-  Wait for a well defined log entry that only occurs when all builds have been finished importing and all belonging diff comparisons have finished calculating (put an appropriate message for that into the application such that we can simply wait for this well defined message)

Implementation details
- BuildImporter schedules a special Job which will execute after all builds are imported and all comparisons are calculated. It writes this log entry "All Builds Imported and Comparisons Calculated."
- CircleCi config was adapted to parse the scenarioo-server.log and wait for this log entry, unless an ERROR log is encountered, in this case the build is aborted.
- Bonus: Fixed ignored tests in ComparisonExecutorTest

fixes #637 
fixes #704


scenarioo-server/src/test/java/org/scenarioo/business/diffViewer/ComparisonExecutorTest.java
```diff
@@ -163,23 +175,19 @@ public void testGetComparisonConfigurationsForBaseBranch2() {
 		assertThat(result.get(1)).isEqualToComparingFieldByField(comparisonConfiguration6);
 	}
 
-	// TODO Fix or delete
-	@Ignore
```
mi-we: 👍

scenarioo-server/src/main/java/org/scenarioo/business/builds/BuildImporter.java
```diff
@@ -120,14 +120,20 @@ public synchronized int submitUnprocessedBuildsForImport(AvailableBuildsList ava
...
 				importNeededBuilds.add(buildImportSummary.getIdentifier());
 				submitBuildForImport(availableBuilds, buildImportSummary.getIdentifier());
 			}
 		}
 
+		//After all Builds were imported and all Comparisons triggered, we want to log a successful completion message.
+		//Since the scheduling of comparisons only  happens after all builds are imported, we need to add a second
+		//Job into the queue to ensure that it is triggered last.
+		asyncBuildImportExecutor.submit(() ->
+			asyncBuildImportExecutor.submit(() -> comparisonExecutor.scheduleComparisonFinishedLog()));
```
mi-we: I would put this in a separate method, it seems less distractive for the flow of the logic (because of the comment)

mi-we: Also, is this correct that the submit happens twice?

eggerschwiler: 1. Done.
2. Yes, because the first submit is triggered after all Builds are imported but before the Comparisons are processed.

The flow looks like this:
 1. All Builds are executed and every Build schedules his Comparisons with the AsyncBuildImporterExecutor to ensure that they are only processed after all Builds have been imported.
 2. The first submit of submitImportFinishedJob is triggered, it schedules another submit, since the Comparisons have not finished yet.
 3. All Comparisons are scheduled in ComparisonExecutor
 4. The second submit of submitImportFinishedJob is triggered, it schedules a submit in ComparisonExecutor.
 5. ComparisonExecutor processes all Comparisons
 6. ComparisonExecutor processes scheduleComparisonFinishedLog.

scenarioo-server/src/test/resources/org/scenarioo/business/diffViewer/build1_build.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
danielsuter: Wouldn't it be better to name the build according to what it is? Like successfull_ build.xml or so?

.circleci/config.yml
```diff
@@ -66,6 +75,7 @@ jobs:
             zip -r ../e2eScenariooDocu.zip ./*
             popd
             curl -f --user scenarioo:justfortest -F"file=@e2eScenariooDocu.zip" http://localhost:8080/scenarioo/rest/builds
+          no_output_timeout: 2m # abort build if no output is produced for 2 minutes
```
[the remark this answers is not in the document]
bruderol: its more than enough - question is: isnt it too much?? If it was enough to wait 30 seconds befroe without any check, I think we should now let it fail after 30 seconds if no output was produced.

scenarioo-server/src/main/java/org/scenarioo/business/builds/BuildImporter.java
```diff
@@ -120,17 +120,37 @@ public synchronized int submitUnprocessedBuildsForImport(AvailableBuildsList ava
...
+	 *
+	 * The flow looks like this:
+	 * 1. All Builds are executed and every Build schedules his Comparisons with the AsyncBuildImporterExecutor to ensure
+	 *    that they are only processed after all Builds have been imported.
+	 * 2. The first submit of submitImportFinishedJob is triggered, it schedules another submit, since the Comparisons have not finished yet.
+	 * 3. All Comparisons are scheduled in ComparisonExecutor
+	 * 4. The second submit of submitImportFinishedJob is triggered, it schedules a submit in ComparisonExecutor.
+	 * 5. ComparisonExecutor processes all Comparisons
+	 * 6. ComparisonExecutor processes scheduleComparisonFinishedLog.
+	 */
```
bruderol: I wouldnt expect a such a long comment on a private method to explain all the outer context and the full flow. I think this can be drastically shortened and/or documented in another place, because this private method only does a small part of all of this.

.circleci/config.yml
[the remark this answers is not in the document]
bruderol: Ich habe das vermutlich falsch verstanden. Bezieht sich ja vermutlich nur auf den Output vom Build Script. Da sind 2 minuten meiner meinung nach ok, da bisher 30 Sekunden genügend waren, oder nicht? sobald es mehr wird, können wir es easy erhöhen.

scenarioo-server/src/test/resources/org/scenarioo/business/diffViewer/build1_build.xml
eggerschwiler: It was a conscious decision to number these files, because there are 3 builds which have a chronological order and all three builds are used in the test.

PR #875, opened 2019-06-21
migrate metaDataTree directive into component

sascha-kaufmann:


scenarioo-client/app/shared/metadata/metaDataTree.component.js
```diff
@@ -15,12 +15,10 @@
...
 
-angular.module('scenarioo.directives').directive('scMetadataTree', function() {
-    return {
-        restrict: 'E',
+angular.module('scenarioo.directives')
+    .component('scMetadataTree', {
         template: require('./metaDataTree.html'),
-        scope: {
+        bindings: {
             metadataTree: '='
```
danielsuter: do we need at two way binding here?

sascha-kaufmann: Nope, changed to one-way binding.

PR #876, opened 2019-06-21
Feature/migrate managed tabs only

danielsuter:


scenarioo-client/app/manage/manage-tabs/manage-tabs.component.ts
```diff
@@ -0,0 +1,58 @@
+import {Component, EventEmitter, Input, OnInit, Output, ViewChild} from '@angular/core';
+import {downgradeComponent} from '@angular/upgrade/static';
+import {TabDirective, TabsetComponent} from 'ngx-bootstrap';
+
+declare var angular: angular.IAngularStatic;
+
+@Component({
+    selector: 'sc-manage-tabs',
+    template: require('./manage-tabs.component.html'),
+    styles: [require('./manage-tabs.component.css').toString()],
```
sascha-kaufmann: Replace with `styleUrls: ['./manage-tabs.component.css']`?

danielsuter: Can't do that because something is wrong with the webpack component. But normal angular would be like this.

PR #884, opened 2019-06-21
Feature/upgrade location service

danielsuter:


scenarioo-client/app/shared/location.service.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Delete after full migration.
+ * This service allows to access the AngularJS location service in Angular components.
+ */
+@Injectable()
+export class LocationService {
+
+    search(queryParameter?: string, value?: string): string | any {
+        return undefined;
```
eggerschwiler: This looks strange to me, maybe add a comment, why we return undefined?

danielsuter: Unfortunately this is the weird API of the AngularJS location service. If you don't provide a parameter, it returns nothing. I can't overload methods in typescript.

eggerschwiler: Thanks for clarifying this.

PR #885, opened 2019-06-21
Feature/migrate diffviewer services

danielsuter:


scenarioo-client/app/shared/location.service.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Delete after full migration.
+ * This service allows to access the AngularJS location service in Angular components.
+ */
+@Injectable()
+export class LocationService {
+
+    search(queryParameter?: string, value?: string): string | any {
+        return undefined;
```
eggerschwiler: This looks strange to me, maybe add a comment, why we return undefined?

eggerschwiler: Moved comment to other pull request where this was introduced.

PR #887, opened 2019-06-24
#886 show error details in case of StepNotFound.

### LINKED ISSUE #886: "The step you are looking for does not exist" is missing information
When a not existing step is encountered in the Scenarioo Viewer the "The step you are looking for does not exist" error page appears.

This error page is missing the following information:
- REST URL to get step
- HTTP method
- HTTP response status
- Response data (was blank on master as well)

StepDoesNotExist

eggerschwiler:
fixes #886 

On the Step Not Found page the error details were not displayed correctly because the error object of HttpClient had changed.

### DIFF
--- a/scenarioo-client/app/step/step.controller.ts
+++ b/scenarioo-client/app/step/step.controller.ts
@@ -141,7 +141,7 @@ function StepController($scope, $routeParams, $location, $route, StepResource, S
                     $scope.stepNotFound = true;
                     $scope.httpResponse = {
                         status: error.status,
-                        method: "GET",
+                        method: 'GET',
                         url: error.url,
                         data: error.error,
                     };
--- a/scenarioo-client/app/step/step.controller.ts
+++ b/scenarioo-client/app/step/step.controller.ts
@@ -141,9 +141,9 @@ function StepController($scope, $routeParams, $location, $route, StepResource, S
                     $scope.stepNotFound = true;
                     $scope.httpResponse = {
                         status: error.status,
-                        method: error.config.method,
-                        url: error.config.url,
-                        data: error.data,
+                        method: "GET",
+                        url: error.url,
+                        data: error.error,
                     };
                 },
             );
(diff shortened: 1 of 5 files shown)


scenarioo-client/app/step/step.controller.ts
```diff
@@ -141,9 +141,9 @@ function StepController($scope, $routeParams, $location, $route, StepResource, S
                     $scope.stepNotFound = true;
                     $scope.httpResponse = {
                         status: error.status,
-                        method: error.config.method,
-                        url: error.config.url,
-                        data: error.data,
+                        method: 'GET',
```
danielsuter: Why can we now fix this?

scenarioo-client/test/e2e/pages/stepPage.ts
```diff
@@ -74,6 +74,11 @@ class StepPage {
         await expect(element(by.id('fallbackMessage')).isDisplayed()).toBeFalsy();
     }
 
+    async assertErrorResponseIsShown() {
+        await expect(element(by.id('stepNotFoundErrorResponseMethod')).getText()).toEqual('GET');
```
danielsuter: I would put the assertion parameters (e.g. GET and 404 as parameters). I know we only use it once but I prefer to have it visible in the webtest itself rather than the page object. So you get some idea what is asserted.

scenarioo-client/test/e2e/specs/step_view.ts
```diff
@@ -72,6 +72,7 @@ useCase('Step - View')
             .it(async () => {
                 await Utils.navigateToRoute('/step/Find Page/find_no_results/inexistent_page.jsp/0/42');
                 await StepPage.assertErrorMessageIsShown();
+                await StepPage.assertErrorResponseIsShown();
```
danielsuter: See comment above.

scenarioo-client/app/step/step.controller.ts
eggerschwiler: This information was not part of the error object. On the other hand, I noticed on line 127 that "StepResource.get" is called, so I'm quite certain that this is always a GET request.

scenarioo-client/test/e2e/pages/stepPage.ts
eggerschwiler: Good idea. I updated this.

Issue #877, opened 2019-06-21
Rest Service providing information if Imports or Comparisons are currently being processed


eggerschwiler:
To ensure that our Demo server deployment does not stop the Tomcat while Imports and Comparisons are being processed, it would be helpful to have a REST service which returns a boolean whether anything is currently being processed.

This should then be checked in the scenarioo-infrastructure reimportBuilds Script.

eggerschwiler:
Has been merged into develop.

PR #890, opened 2019-06-26
#877 Rest Service providing information if Imports or Comparisons are currently being processed

eggerschwiler:
Supply rest endpoint which returns true if all imports were imported and all comparisons were calculated.

Deploy war into a running Tomcat, but wait until all wars are deployed and all imports are finished before restarting the Tomcat. When triggering updateAndImport also wait until all imports are finished.

See corresponding pull request in scenarioo-infrastructure: https://github.com/scenarioo/scenarioo-infrastructure/pull/13

### DIFF
--- a/scenarioo-server/src/main/java/org/scenarioo/business/builds/BuildImporter.java
+++ b/scenarioo-server/src/main/java/org/scenarioo/business/builds/BuildImporter.java
@@ -77,7 +77,7 @@ public class BuildImporter {
 	/**
 	 * Executor to execute one import task after the other asynchronously.
 	 */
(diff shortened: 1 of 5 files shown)


eggerschwiler:
> 
> 
> @eggerschwiler it might be a good idea to do this change on release branch - since we are short before a release and I want to prevent master branch to have conceptual differences to devleop branch for a long time - I think doing such changes now only on develop would mean that same deployment process might not work for release branch and master branch in future.

@bruderol Good point! Unfortunately, Github didn't let me change the source for this pull request (and changing the destination would have also added all the changes that happened in the meantime), thus I close this pull request and open pull request #898 instead.

PR #893, opened 2019-06-27
#892 let spring choose correct content-type

### LINKED ISSUE #892: DiffViewer and Images broken in IE11
Following error in console: DOM7009: Unable to decode image at URL

sascha-kaufmann:
let spring choose correct content-type depending on specified accept headers to fix broken images in IE11

Additional fix for ScreenshotResource.getScreenshotStable method is required, as firefox sends accept headers that spring can not handle properly with produces=image/*

fixes #892

### DIFF
--- a/scenarioo-server/src/main/java/org/scenarioo/rest/step/ScreenshotResource.java
+++ b/scenarioo-server/src/main/java/org/scenarioo/rest/step/ScreenshotResource.java
@@ -50,7 +50,7 @@ public class ScreenshotResource {
 	 * This method is used internally for loading the image of a step. It is the faster method, because it already knows
 	 * the filename of the image.
 	 */
-	@GetMapping(path = "{scenarioName}/image/{imageFileName}", produces = "image/jpeg")
+	@GetMapping(path = "{scenarioName}/image/{imageFileName}", produces = "image/*")
 	public ResponseEntity getScreenshot(@PathVariable("branchName") final String branchName,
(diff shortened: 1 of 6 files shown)


scenarioo-server/src/main/java/org/scenarioo/rest/step/ScreenshotResource.java
```diff
@@ -66,19 +66,39 @@ public ResponseEntity getScreenshot(@PathVariable("branchName") final String bra
...
 
+		return getScreenshotStable(branchName, buildName, usecaseName, scenarioName, pageName, pageOccurrence, stepInPageOccurrence, fallback, labels);
+	}
+
+	/**
+	 * This method is used for sharing screenshot images. It is a bit slower, because the image filename has to be
+	 * resolved first. But it is also more stable, because it uses the new "stable" URL pattern.
+	 * Additional method with produces=image/jpg is needed, as firefox is a bit picky in its accept headers, so produces image/* does not work.
+	 */
+	@GetMapping(path = "{scenarioName}/pageName/{pageName}/pageOccurrence/{pageOccurrence}/stepInPageOccurrence/{stepInPageOccurrence}/image.{extension}", produces = "image/jpg")
```
bruderol: before the content type was "image/jpeg" - and i think this would be more correct - according to https://stackoverflow.com/questions/33692835/is-the-mime-type-image-jpg-the-same-as-image-jpeg

@sascha-kaufmann @eggerschwiler

PR #899, opened 2019-07-01
Bugfix/897 standard jpeg content type

### LINKED ISSUE #897: use standard jpeg content type in resources
before the content type was "image/jpeg" - and i think this would be more correct - according to https://stackoverflow.com/questions/33692835/is-the-mime-type-image-jpg-the-same-as-image-jpeg

@sascha-kaufmann @eggerschwiler

_Originally posted by @bruderol in https://github.com/scenarioo/scenarioo/pull/893/review_comment/create_

eggerschwiler:
The correct content type for jpeg files is "image/jpeg".

Testing with a step that returned a JPEG file, I noticed that even though the Share link worked, the Step View was broken in IE11. Thus some additional changes were needed for the getScreenshot method.

On the demo I changed one step and exchanged 000.png with a converted 000.jpg. This can be tested with the following link (the rest of the steps use regular png files):
http://demo.scenarioo.org/scenarioo-bugfix-897-standard-jpeg-content-type/#/step/Application%20Startup/First%20visit/home/0/0?branch=scenarioo-bugfix-897-standard-jpeg-content-type&build=build-1269

fixes #897

### DIFF
--- a/scenarioo-server/src/main/java/org/scenarioo/rest/step/ScreenshotResource.java
+++ b/scenarioo-server/src/main/java/org/scenarioo/rest/step/ScreenshotResource.java
@@ -100,12 +100,12 @@ public class ScreenshotResource {
 	 * Additional method with produces=image/jpeg is needed, as firefox is a bit picky in its accept headers, so produces image/* does not work.
 	 */
 	@GetMapping(path = "{scenarioName}/pageName/{pageName}/pageOccurrence/{pageOccurrence}/stepInPageOccurrence/{stepInPageOccurrence}/image.{extension}", produces = "image/jpeg")
-	public ResponseEntity getScreenshotStableJpg(@PathVariable("branchName") final String branchName,
-											  @PathVariable("buildName") final String buildName, @PathVariable("usecaseName") final String usecaseName,
-											  @PathVariable("scenarioName") final String scenarioName, @PathVariable("pageName") final String pageName,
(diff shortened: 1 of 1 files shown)


bruderol:
Can we also provide a jpeg image in our generated tests data in one of the scenarios (e.g. an additional Technical test case?) for future tests - or how did you echange the image on the demo to be a jpeg?? 

And I think it would be great to have some test cases to test this.

scenarioo-server/src/main/java/org/scenarioo/rest/step/ScreenshotResource.java
```diff
@@ -50,11 +50,28 @@
...
+										   @PathVariable("buildName") final String buildName, @PathVariable("usecaseName") final String usecaseName,
+										   @PathVariable("scenarioName") final String scenarioName, @PathVariable("imageFileName") final String imageFileName) {
 
+		return getScreenshot(branchName, buildName, usecaseName, scenarioName, imageFileName + ".png");
+	}
+
+	/**
+	 * This method is used internally for loading the image of a step. It is the faster method, because it already knows
+	 * the filename of the image.
+	 * Additional method with produces=image/jpeg is needed, as IE 11 is a bit picky in its accept headers, so produces image/* does not work for jpeg images.
```
bruderol: should we provide the comment also on the other method?
And: is it IE that is picky or Firefox? Because on the other endpoint we talk about Firefox. that is picky with headers.

PR #902, opened 2019-07-03
Feature/extract components from step (WIP)

eggerschwiler:
Step is one huge html-page and one huge controller. Extracting functionality into components will greatly simplify migration and future changes.

The first step was to extract the StepNotFound block.

@danielsuter I had problems to get the binding for stepNotFound in respect with ng-show to work inside my component, so I moved this outside. Could you quickly review these changes, so that I have an accepted baseline for further components?

### DIFF
--- a/scenarioo-client/app/step/stepNotFound/stepNotFound.component.ts
+++ b/scenarioo-client/app/step/stepNotFound/stepNotFound.component.ts
@@ -24,7 +24,7 @@ angular
         controller: StepNotFoundController,
     });
 
-function StepNotFoundController($location, $element) {
+function StepNotFoundController($location) {
     const ctrl = this;
 
     ctrl.getCurrentUrl = getCurrentUrl;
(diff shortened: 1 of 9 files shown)


scenarioo-client/app/step/stepNotFound/stepNotFound.component.ts
```diff
@@ -0,0 +1,35 @@
...
+    .module('scenarioo.directives')
+    .component('scStepNotFoundDiv', {
+        bindings: {
+            response: '<',
+        },
+        template: require('./stepNotFound.html'),
+        controller: StepNotFoundController,
+    });
+
+function StepNotFoundController($location, $element) {
```
danielsuter: $element seems obsolete

eggerschwiler: True. I missed that, thanks!

PR #922, opened 2019-07-05
correct non fast forwarded master

bruderol:
There is a commit on master that is not on release branch and not on develop - this would not be the case if we did that merge with fast-forward. Master should only be fast forwarded.

@eggerschwiler you can call me a nitpicker ;-)

once this is merged please prcoeed as follows:
1. fast forward master to release branch
2. adjust last release tag to point to head of master
3. merge release/5.0 as well into devleop


bruderol:
why was the cricle ci build canceled?

eggerschwiler:
The build was canceled, because the change only affected the docs and publish_docs ran successfully. Since the tag was not changed, the Master Demo would (probably) have received another version number, and I wanted to prevent that, so I aborted the build.

Apparently, I don't understand all the nuances of git yet. I don't see any changes in this pull request but to be sure that we don't run into problems in the future. I'll merge, fast forward master and adjust the tag.

eggerschwiler:
All steps completed. 
- Master was fast-forwarded. 
- Tag moved
- release merged into develop.

PR #945, opened 2019-07-25
Feature/859 new filterable list generated types

eggerschwiler:
Fix typing issue that was encountered when switching to TypeScript because our old code added fields to types that were defined in the backend.

Introduce new types where necessary and map them in the service.

PS: The failing tests also currently fail on the base branch, which is a WIP.


eggerschwiler:
@danielsuter is this an acceptable solution for the hack of adding properties to types defined in the backend?

scenarioo-client/app/diffViewer/types/StepDescription.ts
```diff
@@ -0,0 +1,21 @@
+import {ILabels, IStepDescription} from '../../generated-types/backend-types';
+import {StepDiffInfo} from './StepDiffInfo';
+
+export class StepDescription {
+    details: { [index: string]: any };
```
danielsuter: do we really need to map to any here?

eggerschwiler: This is defined in this way in the backend-types as well:
```
export interface IDetailable {
    details: { [index: string]: any };
}
```

PR #961, opened 2019-09-28
Feautre/migrate branch aliases

danielsuter:
Migration of Branch Aliases. I used template forms because of their simplicity. Probably Reactive forms could be a better fit for such an use-case.
I did not add any test as the E2E Tests cover already everything.


scenarioo-client/app/manage/branch-aliases/branch-aliases.component.ts
```diff
@@ -0,0 +1,89 @@
...
+@Component({
+    selector: 'sc-branch-aliases',
+    template: require('./branch-aliases.component.html'),
+    styles: [require('./branch-aliases.component.css').toString()],
+})
+export class BranchAliasesComponent implements OnInit {
+
+    branchAliases: IBranchAlias[] = [];
+    branches$: Observable<IBranchBuilds[]>;
+    successfullyUpdatedBranchAliases = false;
```
mi-we: How far do we go with declaring the types of variables/return types of methods in TypeScript? Is there some kind of convention e. g. "everything not primitive"?

scenarioo-client/test/e2e/pages/branchAliasesPage.ts
```diff
@@ -35,13 +35,12 @@ class BranchAliasesPage {
         const rowToEditIndex = count - 1;
         const rowToEdit = rows.get(rowToEditIndex);
 
-        await rowToEdit.$('input[name="aliasName"]').sendKeys(name);
+        await rowToEdit.$('input[data-type="branchAlias"]').sendKeys(name);
         if (referencedBranchName !== '') {
```
mi-we: Maybe this method could be split up in two  "enterAlias" and "enterAliasWithoutBranchReference" or something similar. Would make it even clearer to comprehend in the test imho.

scenarioo-client/test/e2e/pages/branchAliasesPage.ts
```diff
@@ -14,7 +14,7 @@ class BranchAliasesPage {
     }
 
     async assertNumberOfAliases(expectedCount) {
-        const rows = this.branchAliasTable.all(by.css('tbody tr'));
+        const rows = this.branchAliasTable.$$('tbody tr');
```
mi-we: In the "enterAlias" method below we do basically the same, but not with the $$. Is there a specific reason for this? Otherwise it should be aligned with this change.

scenarioo-client/app/manage/branch-aliases/branch-aliases.component.ts
danielsuter: I think the linter does it as you suggested. I think it's also valid to specifiy primitive types. Some ppl says if you directly assign it, no type is requiered.

scenarioo-client/test/e2e/pages/branchAliasesPage.ts
danielsuter: done

PR #963, opened 2019-10-04
Feature/new style for detailarea

SteffiSchlumpf:
- Change colors of detailarea
- Resize height of detailarea to page size
- Set first letter in detail-accordion to uppercase


scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.ts
```diff
@@ -182,7 +182,7 @@ export class UseCasesOverviewComponent {
 
     createBranchInformationTree(branch) {
         const branchInformationTree: any = {};
-        branchInformationTree.description = branch.description;
+        branchInformationTree.Description = branch.description;
```
danielsuter: does this really need to be upper case?

scenarioo-client/app/components/detailarea/detail-accordion/detail-accordion.component.css
```diff
@@ -1,7 +1,40 @@
...
+.panel-title {
+    color: #337AB7;
+}
+
+accordion {
+    margin-right: 20px;
+    margin-left: 10px;
+}
+
+accordion accordion-group {
```
danielsuter: the important selector should be avoided if possible. Usually it's only needed if a selector is not specific enough https://uxengineer.com/css-specificity-avoid-important-css/

scenarioo-client/app/components/detailarea/detailarea.component.css
```diff
@@ -37,23 +40,25 @@ sc-detailarea .detail-row {
...
     background-color: #CCCCCC;
     width: 30px !important;
-    vertical-align: middle;
     font-size: x-large;
     color: #337AB7;
-    text-align: center;
     cursor: pointer;
 }
 
+sc-detailarea #arrowCenter {
```
danielsuter: is an id (arrowCenter) really needed or could an additional class achieve the same?

scenarioo-client/app/components/detailarea/detailarea.component.ts
```diff
@@ -53,6 +53,20 @@ export class DetailareaComponent {
...
+    ngOnInit(): void {
+        this.setHeightOfDetailarea();
+    }
+
+    @HostListener('window:resize', ['$event'])
+    onResize() {
+        this.setHeightOfDetailarea();
+    }
+
+    setHeightOfDetailarea() {
```
danielsuter: can this be private or is it called in the template as well?

scenarioo-client/app/components/detailarea/detailarea.component.ts
```diff
@@ -53,6 +53,20 @@ export class DetailareaComponent {
...
+        const headerHeight = document.getElementById('sc-metadata-panel').offsetTop + 2;
+        document.getElementById('sc-metadata-panel').style.height = 'calc(100vh - ' + headerHeight + 'px)';
```
danielsuter: this should be set through a binding not directly in the DOM. E.g. create a member variable metaDataPanelHeight and reference it in the the template. I'm not sure though, if it's possible since the calculation is rather complicated.

danielsuter: To access DOM elements direclty in Angular use:
```import {ElementRef,Renderer2} from '@angular/core';
@ViewChild('someVar') el:ElementRef;

constructor(private rd: Renderer2) {}

ngAfterViewInit() {
      console.log(this.rd); 
      this.el.nativeElement.focus();      //<<<=====same as oldest way
}```

https://stackoverflow.com/questions/38944725/how-to-get-dom-element-in-angular-2/39679241

File: scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.ts (hunk 1 above)
Comment (SteffiSchlumpf) [reply scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.ts]: If it's not upper case here, is is written in lower case in the detail area:
detailarea

Would it be better to set
`nodeLabel: key.charAt(0).toUpperCase() + key.slice(1),` 
on line 64 in treeDataCreator.pipe.ts?

File: scenarioo-client/app/components/detailarea/detailarea.component.css (hunk 3 above)
Comment (SteffiSchlumpf) [reply scenarioo-client/app/components/detailarea/detailarea.component.css]: An additional class could achieve the same. 
But is it not better to use an id in this case? It appears only once...

File: scenarioo-client/app/components/detailarea/detailarea.component.ts (hunk 4 above)
Comment (SteffiSchlumpf) [reply scenarioo-client/app/components/detailarea/detailarea.component.ts]: it can be private

File: scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.ts (hunk 1 above)
Comment (bruderol) [reply scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.ts]: no no this should be done only in the view when outputting the keys and not in the data structure that is passed - so the details accordion itself should take care of it, no matter what keys are passed to it to always display the names of the keys in a human readable way (spaces and upper case first letter)

File: scenarioo-client/app/components/detailarea/detailarea.component.css (hunk 3 above)
Comment (bruderol) [reply scenarioo-client/app/components/detailarea/detailarea.component.css]: an id, is something that needs to be unique on the whole page. I would only use this in styling, if it is for sure something that will never never never occur twice on one page. So I would not use an id here, you never know, one day there could be two detail areas on the same page, couldnt it?

File: scenarioo-client/app/components/detailarea/detailarea.component.ts (hunk 5 above)
Comment (bruderol) [reply scenarioo-client/app/components/detailarea/detailarea.component.ts]: could the same not simply be achieved by setting minHeight to 100% for the details area?

File: scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.ts (hunk 1 above)
Comment (bruderol) [reply scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.ts]: please handle this inisde the component that renders these titles and not whewre the data is passed to it. We have other places where lowercase keys will be passed to accordions that need to be handled in general by the component in one place.

File: scenarioo-client/app/components/detailarea/detailarea.component.css (hunk 3 above)
Comment (bruderol) [reply scenarioo-client/app/components/detailarea/detailarea.component.css]: See also https://dev.to/clairecodes/reasons-not-to-use-ids-in-css-4ni4

Comment (SteffiSchlumpf) [reply scenarioo-client/app/components/detailarea/detailarea.component.css]: All right, thanks.

File: scenarioo-client/app/components/detailarea/detailarea.component.ts (hunk 5 above)
Comment (SteffiSchlumpf) [reply scenarioo-client/app/components/detailarea/detailarea.component.ts]: > could the same not simply be achieved by setting minHeight to 100% for the details area?

This is unfortunately not possible. By setting minHeight to 100%, it doesn't subtract the space above the detail area.

File: scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.ts (hunk 1 above)
Comment (SteffiSchlumpf) [reply scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.ts]: @danielsuter: I looked at it with Rolf and we decided to keep it that way. Whoever writes node labels by hand is also responsible for whether this should be written in upper or lower case.

File: scenarioo-client/app/pipes/treeDataCreator.pipe.ts [hunk 6]
```diff
@@ -51,17 +51,17 @@ export class TreeDataCreatorPipe implements PipeTransform {
             const value = node[key];
             if (Array.isArray(value)) {
                 childNodes.push({
-                    nodeLabel: key,
+                    nodeLabel: key.charAt(0).toUpperCase() + key.slice(1),
```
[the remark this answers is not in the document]
SteffiSchlumpf: That's not the current state (outdated). We have decided to keep it as it was before:
`nodeLabel: key`
Whoever writes node labels by hand is also responsible for whether this should be written in upper or lower case.

scenarioo-client/app/build/scenarios-overview/scenarios-overview.component.html
```diff
@@ -15,8 +15,7 @@
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 
-<sc-detailarea (togglePannelCollapsedValue)="collapsePanel($event)" [mainDetailsSections]="mainDetailsSections"
-               [additionalDetailsSections]="additionalDetailsSections" key="useCaseView">
```
bruderol: cool - now it looks very clean :-D

scenarioo-client/app/build/scenarios-overview/scenarios-overview.component.html
```diff
@@ -15,133 +15,125 @@
...
-                </span>
-            </label>
-        </div>
+            </span>
+        </label>
     </div>
 
-    <div id="sc-meta-data-panel">
-        <sc-detailarea (togglePannelCollapsedValue)="collapsePanel($event)" [mainDetailsSections]="mainDetailsSections" [additionalDetailsSections]="additionalDetailsSections" key="useCaseView"></sc-detailarea>
+    <p class="tableTitle">
```
bruderol: I would use div instead of `<p>` elements - or what is the reason for the `<p>` elem here??

scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.html
```diff
@@ -15,8 +15,9 @@
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 
-<div class="row sc-space-top-large sc-space-bottom">
-    <div class="col-sm-8">
+<sc-detailarea [mainDetailsSections]="mainDetailsSections" key="mainView">
+
+    <div class="searchbar">
```
bruderol: I think this all looks so much cleaner now :-) :+1:

scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.ts
```diff
@@ -182,10 +177,6 @@ export class UseCasesOverviewComponent {
         }
     }
 
-    collapsePanel(isPanelCollapsed: boolean) {
-        this.isPanelCollapsed = isPanelCollapsed;
-    }
-
```
bruderol: cool that we got rid of all this extra code on every page :+1: 

I like in what direction this is all going :-)

scenarioo-client/app/components/detailarea/detailarea.component.html
```diff
@@ -15,23 +15,32 @@
...
-    <div id="sc-metadata-panel" class="detail-container col-sm-4 pull-right">
-        <div class="detail-row">
+    <button (click)="togglePannelCollapsedValue()" [attr.aria-expanded]="!isPanelCollapsed" aria-controls="collapseBasic"
+            class="btn btn-primary pull-right" id="sc-showHideDetailsButton" type="button">
+        {{ isPanelCollapsed ? "Show details" : "Hide details" }}
+    </button>
+
+    <div [ngClass]="isPanelCollapsed ? 'col-sm-11' : 'col-sm-8'">
+        <ng-content></ng-content>
+    </div>
```
bruderol: just awesome - look how cool that the detailsarea finally fully controls how to do the outer layout depending on collapsed state :-) 👍

scenarioo-client/app/scenario/scenario.html
```diff
@@ -30,7 +30,7 @@
             </span>
         </div>
     </div>
-    <div class="col-sm-4">
+    <div class="col-sm-2 pull-right">
```
bruderol: was this change really intended?

bruderol: it is just a question - maybe you can explain it to me.

bruderol:
I also reviewed the new style in the app - I think this is now just awesome! :-) 👍

scenarioo-client/app/build/scenarios-overview/scenarios-overview.component.html
SteffiSchlumpf: done

scenarioo-client/app/scenario/scenario.html
SteffiSchlumpf: Yes it was intended to keep the two buttons (expand/collapse all & Hide/Showdetails) in the upper right corner. The buttons were moved because of the search bar.

PR #974, opened 2019-11-07
Transparent color of the active row in the table

SteffiSchlumpf:
Make the color of the active row transparent, so the comparison color is visible.

### DIFF
--- a/scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.css
+++ b/scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.css
@@ -23,9 +23,13 @@ sc-usecases-overview .table thead span {
     cursor: pointer;
 }
 
-sc-usecases-overview .table tbody>tr:hover {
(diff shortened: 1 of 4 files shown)


scenarioo-client/app/build/use-cases-overview/use-cases-overview.component.css
```diff
@@ -23,9 +23,13 @@ sc-usecases-overview .table thead span {
     cursor: pointer;
 }
 
-sc-usecases-overview .table tbody>tr:hover {
+sc-usecases-overview .table tbody>tr:hover td {
```
bruderol: I think it would be cool to encapsulate our tables in our own Filterabel-Table component - such that in future we do not have to duplictae these styles, and logic and so on for tables on serveral pages. Maybe a good exercise for next few weeks for a more complex custom component?

bruderol:
@SteffiSchlumpf you can merge this one!

PR #975, opened 2019-11-07
#973 Do not show arrow outside of the detailarea

### LINKED ISSUE #973: Set height of detailarea equal to the height of the content
image

If you have to scroll to see the whole table, the height of the detailarea stops in the middle of the page, but the arrow is still in the middle of the page.

SteffiSchlumpf:
Fixed #973

### DIFF
--- a/scenarioo-client/app/components/detailarea/detailarea.component.css
+++ b/scenarioo-client/app/components/detailarea/detailarea.component.css
@@ -50,12 +50,14 @@ sc-detailarea .detail-button {
     font-size: x-large;
     color: #337AB7;
     cursor: pointer;
+    vertical-align: top;
 }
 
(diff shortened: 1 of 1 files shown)


bruderol:
I think it is still not optimal - maybe we can have a look on it together on Friday ?

I would rather prefer that the gray area expands to the whole page height - such that the user can still collapse after scrolling all the way down, without having to scroll up again. And that the arrow is placed in the center of the visible area.

bruderol:
Thanks @SteffiSchlumpf - I now overworked this part and refactored the whole page layout of the detailarea part to use flex box instead and also to ensure the sticky thing (inspired from your changes) is   more or less centered independent on sizes of content.

I open another PR for this.

PR #984, opened 2019-11-15
#978 Introduce Swagger and Swagger UI

### LINKED ISSUE #978: Introduce Swagger UI in Spring Backend

bruderol:
I integrated Swagger (API Docs) and Swagger UI, such that REST APIs can be browsed and tested directly on Scenarioo by going to this URL:

.../scenarioo/swagger-ui.html

@SteffiSchlumpf can you have a short look at it and try it out - might be helpful to finalize your image stuff and test REST calls in general.

The UI shows you all our REST calls and when you open one you can "Try it out" through the UI:

image

Fixes #978

### DIFF
--- a/scenarioo-server/build.gradle
+++ b/scenarioo-server/build.gradle
@@ -69,7 +69,7 @@ generateTypeScript {
 }
 
 dependencies {
-    compile 'com.google.guava:guava-base'
+
     compile 'org.apache.xmlgraphics:batik-transcoder'
     compile 'org.apache.xmlgraphics:batik-codec'
     compile 'org.apache.xmlgraphics:xmlgraphics-commons'
@@ -81,6 +81,11 @@ dependencies {
 
     providedRuntime 'org.springframework.boot:spring-boot-starter-tomcat'
 
+
+    compile 'io.springfox:springfox-swagger2:2.9.2'
+    compile 'io.springfox:springfox-swagger-ui:2.9.2'
+    
+    
     //These libraries are no longer provided in Java 9+, thus we need to add them manually
     compile 'javax.xml.bind:jaxb-api'
     compile 'com.sun.xml.bind:jaxb-core'
@@ -148,7 +153,7 @@ bootWar.doLast {
     }
 }
 
-bootWar.dependsOn createVersionPropertiesFile
+assemble.dependsOn createVersionPropertiesFile
 war.dependsOn createVersionPropertiesFile
 bootWar.dependsOn ":scenarioo-client:build"
 war.dependsOn ":scenarioo-client:build"
--- a/scenarioo-server/build.gradle
+++ b/scenarioo-server/build.gradle
@@ -81,10 +81,8 @@ dependencies {
 
     providedRuntime 'org.springframework.boot:spring-boot-starter-tomcat'
 
-
-    compile 'io.springfox:springfox-swagger2:2.9.2'
-    compile 'io.springfox:springfox-swagger-ui:2.9.2'
-    
+    compile 'io.springfox:springfox-swagger2'
+    compile 'io.springfox:springfox-swagger-ui'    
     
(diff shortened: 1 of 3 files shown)


scenarioo-server/build.gradle
```diff
@@ -81,6 +81,11 @@ dependencies {
 
     providedRuntime 'org.springframework.boot:spring-boot-starter-tomcat'
 
+
+    compile 'io.springfox:springfox-swagger2:2.9.2'
```
danielsuter: do we need an explicit version?

scenarioo-server/src/main/java/org/scenarioo/configuration/SwaggerConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import springfox.documentation.builders.PathSelectors;
+import springfox.documentation.builders.RequestHandlerSelectors;
+import springfox.documentation.spi.DocumentationType;
+import springfox.documentation.spring.web.plugins.Docket;
+import springfox.documentation.swagger2.annotations.EnableSwagger2;
+
+@Configuration
```
danielsuter, line 11: this will enable it for everyone correct? Including all customers running scenarioo. Is this intended?

scenarioo-server/build.gradle
bruderol: Of course - versions should always be explicit - we can discuss if you not agree

the question is actually: arent our other dependencies fixed versions? I think they should be!

I rather prefer manually updating versions in a well controlled manual fashion than taking the risk that somebody gets blocked because a new version brings in troubles --> non-deterministic builds (see also the "R" in CRISP build criterias!).

scenarioo-server/src/main/java/org/scenarioo/configuration/SwaggerConfiguration.java
bruderol: I dont see a problem with enabling this for all customers. But if you think its a problem we can make it customizable through a profile easily. But by default I would like to have it enabled. I do not see any problem with it.

scenarioo-server/build.gradle
danielsuter: the versions are fixed through bom files. If we specify a version explicitely it can break everything. That's the whole idea of the bom.

bruderol: remark: most of our other dependency versions are by the way as well fixed -see gradle.build file in the root of scenarioo that specifies these other versions - I can put this version here as well there - to have all in the same place.

bruderol: bom file?? we use gradle not maven, right?

bruderol: ahm I understand, ok, I see - yes would be a good idea, but we do not use BOM yet - is only supported since gradle 5 - we still use gradle 4. So for now we fix all our versions. Its already like that (just in the other file with the dependencies - I think it is better to have them well defined and update them manually from time to time - automatic update almost like always fails anyway!)

scenarioo-server/src/main/java/org/scenarioo/configuration/SwaggerConfiguration.java
davidlahner: Good point. We could make it configurable. Docket has an enable(boolean)-Method and we could use a value configured in application.properties to enable/disable Swagger.

davidlahner (approved):
Looks good. Thanks!

scenarioo-server/build.gradle
bruderol: I moved the version now to the dependency management part in the root gradle.build file, but it's still a fixed version, cause the current used dependency apüproach does not use proper BOM mechanism.

PR #991, opened 2019-12-20
Update gradle wrapper to version 5.6.4.

danielsuter:


danielsuter:
@eggerschwiler Do you have any ideas why the tests are failing? If I deploy locally they fail as well. The test data seems to be different. I don't get why.

davidlahner (approved):
Looks good. Thanks for the initiative!

danielsuter:
> Looks good. Thanks for the initiative!

Thanks for fixing the test issues! My solution would have been much uglier.

PR #997, opened 2020-01-28
Stabilize CalculationStatus

### LINKED ISSUE #985: Bug & Flaky Test: Concurrency problem with ComparisonsResource
I identified that sometimes the test ComparisonsResourceIntegrationTest is failing because of a problem/bug in ComparisonResource.

The problem is: the file that is read to get the comparison status (on endpoint "/calculationStatus") might be written in parallel in some of the test cases (e.g. "recalculate_should_queue_existing_comparison" which only queues a comparison for calculation and then when trying to get the status the background job might at the same time write the file in parallel --> file content is incomplete and can not be parsed properly in that case) - which leads to an error in the service and an unexpected status code in the rest call.

davidlahner:
I adapted the code to return PROCESSING if the buildDiffInfo file cannot be parsed. I hope this should stabilize the test.

Closes #985


bruderol:
@eggerschwiler vielen Dank David! :+1:

PR #1007, opened 2020-02-11
migrate title from controller to component

JonasMa:


.idea/misc.xml
```diff
@@ -1,12 +1,10 @@
...
 <project version="4">
+  <component name="ExternalStorageConfigurationManager" enabled="true" />
   <component name="JavaScriptSettings">
     <option name="languageLevel" value="ES6" />
   </component>
-  <component name="ProjectPlainTextFileTypeManager">
-    <file url="file://$PROJECT_DIR$/../scenarioo.github.io/_includes/css/main.css" />
-  </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="13" project-jdk-type="JavaSDK">
```
StuckiSimon, line 7: Have you changed the settings by accident or on purpose?

scenarioo-client/app/shared/navigation/title.component.ts
```diff
@@ -0,0 +1,22 @@
+import * as angular from 'angular';
+import {ConfigurationService} from '../../services/configuration.service';
+import {Title} from '@angular/platform-browser';
+import {Component, Directive} from '@angular/core';
+import {downgradeComponent} from '@angular/upgrade/static';
+
+@Component({
+    selector: 'app-title',
+    template: '',
+})
```
StuckiSimon, line 10: Have you considered using a service? (Since there is no actual HTML involved)

StuckiSimon (commented):
Looks solid, but maybe using a service would serve the purpose? Is there any benefit of using a component / directive here?

.idea/misc.xml
JonasMa: accident

scenarioo-client/app/shared/navigation/title.component.ts
JonasMa: don't you like having it on the index.html?

JonasMa:
It would make sense to inject the Angular title service in the navigation.controller.ts but unfortunately this is still in AngularJS. When this file is migrated it's the right time (and easiest) to implement this change. Currently this wouldn't introduce any benefits

PR #1009, opened 2020-02-11
Feature/embed uml diagrams for better tool support

bruderol:
And another change regarding UML diagrams - it seems intelliJ has better support for embedded uml code than external files - the advantage: you see it directly in the markdown editor preview (which does not work somehow for external files, sadly)


bruderol:
ich merge das mal - hoffe ist ok?

PR #1013, opened 2020-02-11
Detailarea Improvements / Cleanup

bruderol:
1. detailarea to take full height of viewport AND of page (if right content is higher than viewport)
2. detail area collapse arrow not to slide out of the arrow button box
3. detail area collapse arrow to be as much centered as possible
4. no table in layout css - use flex box instead
5. clean up some namings (no metadata anymore etc.)
6. clean up some css styles that should not be needed

### DIFF
--- a/scenarioo-client/app/components/detailarea/detailarea.component.css
+++ b/scenarioo-client/app/components/detailarea/detailarea.component.css
@@ -1,5 +1,10 @@
-/* Buttons over detailarea */
+/* Details Area Container (top level div) */
+sc-detailarea .details-area-container {
+    margin-top: 35px;
+    height: 100%;
+}
 
+/* Buttons over detailarea (to hide / expand the details area) */
 sc-detailarea button.btn.btn-primary {
     background-color: unset;
     border: unset;
@@ -20,42 +25,57 @@ sc-detailarea button.btn.btn-primary:active {
     color: #333333;
 }
 
+/* General flex box layout styles */
 
-/* Detailarea */
-
-sc-detailarea .sc-meta-data-panel {
-    margin-top: 35px;
+sc-detailarea .flex {
+    display: flex;
 }
 
-sc-detailarea .detail-container {
+sc-detailarea .flex-fill {
+    flex: 1 1 auto;
+}
+
+/* Colapsable Details Panel */
+
+sc-detailarea .details-panel {
     border: 1px solid #ccc;
     border-right: 0;
-    display: table;
     padding: unset;
     background-color: #eeeeee;
-    right: 0;
 }
 
-sc-detailarea .detail-row {
-    display: table-row;
-}
-
-sc-detailarea .detail-cell {
-    display: table-cell;
-}
-
-sc-detailarea .detail-button {
+sc-detailarea .details-collapse-button {
     background-color: #CCCCCC;
     width: 30px;
+    min-width: 30px; /* do never shrink */
     font-size: x-large;
     color: #337AB7;
     cursor: pointer;
 }
 
+sc-detailarea .sticky-arrow-container {
+    display: block;
+    vertical-align: top;
+    top: 0;
+    height: 100%;
+}
+
+sc-detailarea .sticky-arrow-icon-area {
+    position: -webkit-sticky;
+    position: sticky;
+    top: 33vh; /* 2 times this plus the height should add to 100vh => sticky arrow is centered once full container is scrolled into view */
+    height: 34vh;
+    /* reasons for choosen height - it seems a good compromise between:
+     * 100vh = having to scroll too far down on small screens until you see the arrow
+     * 5vh = having the arrow too near to the top of the details panel until full panel scrolled into view
+     */
+    vertical-align: middle;
+}
+
 sc-detailarea .arrowCenter {
-    position: fixed;
+    position: relative;
     margin-left: 10px;
-    top: 55%;
+    height: 100%;
 }
 
 sc-detailarea .detail-section {
--- a/scenarioo-client/app/components/detailarea/detailarea.component.css
+++ b/scenarioo-client/app/components/detailarea/detailarea.component.css
@@ -6,8 +6,8 @@ sc-detailarea .details-area-container {
 
 /* Buttons over detailarea (to hide / expand the details area) */
 sc-detailarea button.btn.btn-primary {
-    background-color: unset;
-    border: unset;
+    background-color: transparent;
+    border: none;
     color: #337AB7;
     font-style: italic;
     padding: 0;
@@ -16,9 +16,8 @@ sc-detailarea button.btn.btn-primary {
 }
 
 sc-detailarea button.btn.btn-primary:focus {
-    outline: unset;
     background-color: transparent;
-    box-shadow: unset;
+    box-shadow: none;
 }
 
 sc-detailarea button.btn.btn-primary:active {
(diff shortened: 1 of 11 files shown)


scenarioo-client/app/components/detailarea/detailarea.component.html
```diff
@@ -15,41 +15,83 @@
...
-                    <i class="fas arrowCenter fa-angle-left"></i>
-                </span>
-            </div>
+    <!-- Content Area -->
+    <div class="row flex">
+
+
+        <!-- Page Content left (will fill if details are collapsed) -->
+        <div class="detailarea-page-content"
+             [ngClass]="isPanelCollapsed ? 'col-sm-11 flex-fill' : 'col-sm-8'">
```
lynparbone: As an alternative to _'col-sm-11 flex-fill'_: _col-sm-12_, this causes the detailarea to take up the whole width (bootstrap's grid system has 12 columns)

lynparbone, line 39: Perhaps there is a possiblity to use either flexbox or bootstrap (not both)?

E.g. the full width could be taken up by setting _col-sm-12_ as well (instead of _col-sm-11 flex-fill_).

scenarioo-client/app/components/detailarea/detailarea.component.css
```diff
@@ -20,42 +25,57 @@ sc-detailarea button.btn.btn-primary:active {
...
 
-sc-detailarea .detail-cell {
-    display: table-cell;
-}
-
-sc-detailarea .detail-button {
+sc-detailarea .details-collapse-button {
     background-color: #CCCCCC;
     width: 30px;
+    min-width: 30px; /* do never shrink */
```
StuckiSimon, line 40: Consider using `flex-shrink: 0;` - this would make the comment obsolete.

scenarioo-client/app/components/detailarea/detailarea.component.css
```diff
@@ -20,42 +25,57 @@ sc-detailarea button.btn.btn-primary:active {
...
     font-size: x-large;
     color: #337AB7;
     cursor: pointer;
 }
 
+sc-detailarea .sticky-arrow-container {
+    display: block;
+    vertical-align: top;
+    top: 0;
```
StuckiSimon: Are these properties required?

scenarioo-client/app/components/detailarea/detailarea.component.css
```diff
@@ -20,42 +25,57 @@ sc-detailarea button.btn.btn-primary:active {
...
+sc-detailarea .sticky-arrow-icon-area {
+    position: -webkit-sticky;
+    position: sticky;
+    top: 33vh; /* 2 times this plus the height should add to 100vh => sticky arrow is centered once full container is scrolled into view */
+    height: 34vh;
+    /* reasons for choosen height - it seems a good compromise between:
+     * 100vh = having to scroll too far down on small screens until you see the arrow
+     * 5vh = having the arrow too near to the top of the details panel until full panel scrolled into view
+     */
+    vertical-align: middle;
```
StuckiSimon: Is this property required?

scenarioo-client/app/components/detailarea/detailarea.component.css
```diff
@@ -20,42 +25,57 @@ sc-detailarea button.btn.btn-primary:active {
...
+
+sc-detailarea .sticky-arrow-icon-area {
+    position: -webkit-sticky;
+    position: sticky;
+    top: 33vh; /* 2 times this plus the height should add to 100vh => sticky arrow is centered once full container is scrolled into view */
+    height: 34vh;
+    /* reasons for choosen height - it seems a good compromise between:
+     * 100vh = having to scroll too far down on small screens until you see the arrow
+     * 5vh = having the arrow too near to the top of the details panel until full panel scrolled into view
+     */
```
StuckiSimon: What about something like this:
`position: fixed;`
`top: 50vh;`

StuckiSimon (commented):
Looks fine, there is some CSS which is probably not required.

scenarioo-client/app/components/detailarea/detailarea.component.html
bruderol: Yes, I tried that - but then it looked ugly - it's why this seemed like a good compromise for now - at least they seem to work well together. I see no troubles currently. I have even seen a stackoverflow post were they proposed some similar mixes in bootstrap 4 ("col-sm-11" with "flex-grow-1" - unfortunately we do not have bootstrap 4 yet, but at least this is already almost similar. Maybe we should go for bootstrap 4 ??

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: yep, nice., I'll try that

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: not sure - I was really desperate here - the existing CSS was sooo bad already.

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: I'll check

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: works like a charm - thanks for spotting that!

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: Unbelievable - well spotted - those are really not needed!

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: nope, this is not what I want - the problem here is: with "fixed" it can scroll out of the container box - with sticky it doesnt.

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: not sure why this is not needed anymore - but does it hurt to keep it??
I think it is important this is aligned centered in here - do not understand why it is already without it.

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: but i improved the positioning again - now that the size of the panel is at least the whole page

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: ok, changed this to flex layout now

bruderol:
Ok, reviewed and improved it with @StuckiSimon - Thanks for help! Will merge that now.

bruderol:
I mean now ;-)

bruderol:
@StuckiSimon ok, improved some further details regarding top level page layout and styling the button to show or hide details. will merge it now.

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: was resolved

scenarioo-client/app/components/detailarea/detailarea.component.css
bruderol: was fixed

PR #1014, opened 2020-02-12
#1002 Breadcrumb Migration from AngularJS to Angular

### LINKED ISSUE #1002: Breadcrumb Migration from AngularJS to Angular
Steps:

Service can be migrated completely without wrapper (as it is only used in the breadcrumb directive)
AngularJS Directive needs to be migrated to Angular component
Component needs to support AngularJS
Migration to TypeScript in general

Part of Epic: #1000

StuckiSimon:
Migrate breadcrumbs.
Drop duplicate SharePopup implementation.

Closes #1002


scenarioo-client/app/build/mainpage/mainpage.component.html
```diff
@@ -14,8 +14,10 @@
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
-
-<sc-share></sc-share>
+<div id="grayBackground"></div>
+<span class="button-right">
+    <sc-share></sc-share>
+</span>
```
bruderol, line 20: I dont get why this was necessary

scenarioo-client/test/spec/shared/navigation/sharePage/sharePage.service.spec.ts
```diff
@@ -17,45 +17,59 @@
...
+                expect(value).toBeUndefined();
+                resolve();
+            });
+        });
+        await new Promise((resolve) => {
+            sharePageService.getImageUrl().subscribe((value) => {
+                expect(value).toBeUndefined();
+                resolve();
+            });
+        });
```
bruderol: I dont get why it was necessary to make this service return observables in this case - is it realy necessary ??

scenarioo-client/app/build/mainpage/share/share.component.ts
```diff
@@ -49,21 +49,24 @@ export class ShareComponent implements OnInit {
...
-        this.pageUrl = this.getPageUrl();
+        this.sharePageService.getPageUrl().subscribe((pageUrl) => {
+            this.pageUrl = this.getPageUrl(pageUrl);
+            this.eMailUrl = encodeURIComponent(this.pageUrl);
+        });
 
-        this.imageUrl = this.pageURLService.getImageUrl();
+        this.sharePageService.getImageUrl().subscribe((imageUrl) => {
+            this.imageUrl = imageUrl;
+        });
```
bruderol: I dont get why this had to be observables - why dont we instead only get the values when we really need it - which is when the modal is opened instead of observing them all the times here and storing state in the component that is not needed otherwise.

bruderol: I would recommend to make the used service simpler again (remove observable complexity) and move all the getting values code into the openModal function instead to keep this code simple here. While the popup is open the urls wont change anyways - So I really see no need to make this more complex than needed here.

scenarioo-client/test/spec/shared/navigation/sharePage/sharePage.service.spec.ts
bruderol: see other comment - I really think this is overkill here

bruderol (changes requested):
See my comments - I think some observables should be avoided where not necessary.

Furthermore there is a bug now the the sketcher button in the breadcrumbs bar partialy appears on pages where it did not appear before:
image
(should only appear on step page)

scenarioo-client/app/search/search.html
```diff
@@ -15,7 +15,7 @@
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 
-<sc-breadcrumbs></sc-breadcrumbs>
+<sc-breadcrumbs breadcrumb-id="search"></sc-breadcrumbs>
```
bruderol, line 18: by the way: I like that this has been moved out of the router :+1:

scenarioo-client/app/shared/navigation/breadcrumbs/sketcherLink.service.ts
```diff
@@ -15,45 +15,65 @@
...
+    hideCreateOrEditSketchLinkInBreadcrumbs() {
+        return undefined;
+    }
+
+    isShowCreateOrEditSketchLink() {
+        return undefined;
+    }
+
+    getCreateOrEditSketchLinkTitle() {
+
```
bruderol, line 67: couldnt we make the service a real new angular service class and then downgrade it - instead of the other way round ??? I do not like the fact that we start to produce this half migrated services with this ugly classes that do not do anything. I think at least for this simple service here it would have been possible.

bruderol:
The rest looks great to me - thanks!

scenarioo-client/app/shared/navigation/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,90 @@
...
+                isLast: breadcrumbItem.isLastNavigationElement,
+            };
+        });
+    }
+
+    private getShortenedLabelText(breadcrumbItem, isLabelTextShortened) {
+        return isLabelTextShortened ? this.getShortenedText(breadcrumbItem.label) : breadcrumbItem.label;
+    }
+
+    private getShortenedText(text) {
```
danielsuter: add typing

scenarioo-client/app/shared/navigation/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,90 @@
...
+                text: breadcrumbLabelText,
+                tooltip: breadcrumbItem.textForTooltip,
+                showTooltip: isLabelTextShortened,
+                href: `#${breadcrumbItem.route}`,
+                isLast: breadcrumbItem.isLastNavigationElement,
+            };
+        });
+    }
+
+    private getShortenedLabelText(breadcrumbItem, isLabelTextShortened) {
```
danielsuter: add typing

scenarioo-client/app/shared/navigation/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    private getShortenedText(text) {
+        if (text.length > this.limit) {
+            const shortenedText = text.substr(0, this.limit);
+            return shortenedText + '...';
+        }
+        return text;
+    }
+
+    private getNavigationParameters() {
```
danielsuter: add typing

scenarioo-client/app/shared/navigation/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,90 @@
...
+
+declare var angular: angular.IAngularStatic;
+
+@Component({
+    selector: 'sc-breadcrumbs',
+    template: require('./breadcrumbs.component.html'),
+})
+export class Breadcrumbs implements OnInit {
+    limit = 50;
+    breadcrumbs: any[];
```
danielsuter: create a type

scenarioo-client/app/shared/navigation/breadcrumbs/breadcrumbs.service.ts
```diff
@@ -0,0 +1,188 @@
...
+    search: {
+        breadcrumbPath: [homeElement, searchElement],
+    },
+};
+
+@Injectable()
+export class BreadcrumbsService {
+    constructor(private humanReadable: HumanReadablePipe) {
+    }
+    public getNavigationElements(breadcrumbId, navParameters): NavigationElement[] {
```
danielsuter, line 124: newline use formatting

scenarioo-client/app/shared/navigation/breadcrumbs/breadcrumbs.service.ts
```diff
@@ -0,0 +1,188 @@
...
+                }
+                navigationElement.label = this.setValuesInLabel(navigationElement.label, navParameters);
+                navigationElement.textForTooltip = this.convertToPlainText(navigationElement.label);
+                return navigationElement;
+            });
+        }
+        return [];
+    }
+
+    private convertToPlainText(text) {
```
danielsuter: typing

scenarioo-client/app/shared/navigation/sharePage/sharePage.service.ts
```diff
@@ -0,0 +1,41 @@
...
+declare var angular: angular.IAngularStatic;
+
+@Injectable()
+export class SharePageService {
+
+    private readonly pageUrl: ReplaySubject<string>;
+    private readonly imageUrl: ReplaySubject<string>;
+
+    constructor() {
+        this.pageUrl = new ReplaySubject(1);
```
danielsuter: move to fields

StuckiSimon:
> Furthermore there is a bug now the the sketcher button in the breadcrumbs bar partialy appears on pages where it did not appear before:
> image
> (should only appear on step page)

Good catch, that was due to styling overwrites. Is fixed now 👍

bruderol:
@StuckiSimon see #1015 to reduce complexity here. Thanks

PR #1018, opened 2020-02-14
#1002 Clean up share this page component with its service

bruderol:
* renaming
* file moved: keep things together that belong together
* reduce custom css styles and move to components they belong to
* cleanup warnings with obviously wrong typings


StuckiSimon (approved):

scenarioo-client/app/build/mainpage/mainpage.component.css
```diff
@@ -2,3 +2,10 @@
     border: 0;
     margin-bottom: 20px;
 }
+
+#grayBackground {
```
lynparbone: Maybe _#tabBackground_ instead? I think _grayBackground_ is a bit ambiguous, like it is used only to set a gray background. But it contains additional stylings specific to the tab background.

bruderol: @lynparbone yeah, this was an old naming, I also thought it is a bad name, but was not so sure what would be a better name.
Maybe "tabHeader" because indead it is before the tabs, and not really the tabs?

lynparbone: Yes, I think _tabHeader_ would be suitable as well!

lynparbone (approved):

PR #1030, opened 2020-11-04
Utility CSS styles for flexbox

bruderol:
As discussed, due to the fact, that we have currently no bootstrap 4


SteffiSchlumpf:
finde ich gut 👍

PR #1034, opened 2020-12-03
Fixes chrome version

danielsuter:


davidlahner (changes requested):
in circle-ci/config.yml there is still the old image used on line 8 (circleci/openjdk:8-jdk-browsers), is this on purpose?

danielsuter:
@eggerschwiler Good catch, that was a merge fail.

bruderol (approved):
top 👍 Danke!!

bruderol:
@eggerschwiler I guess this can be merged? what you think?

davidlahner (approved):

PR #1038, opened 2020-12-03
Feature/caching parallel execution

danielsuter:
Biggest bottle neck is elastic search (9 minutes only circleci). So this will "only" improve local dev atm.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,21 +20,21 @@ jobs:
           name: Clean & install
           command: |
             mkdir -p test-results
-            ./gradlew clean build -x test --no-daemon -s
+            ./gradlew clean build -x test -s
       - run:
           name: Unit tests
-          command: ./gradlew test -x scenarioo-client:npm_test --continue --no-daemon -s --info
+          command: ./gradlew test -x scenarioo-client:npm_test --continue -s --info
       - run:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -20,21 +20,21 @@ jobs:
           name: Clean & install
           command: |
             mkdir -p test-results
-            ./gradlew clean build -x test --no-daemon -s
+            ./gradlew clean build -x test -s
```
bruderol, line 23: bisch sicher das das  e gueti idee isch? isch unsere build wirklich safe mit daemon?? ich persönlich ha lieber e reliable CI build wo nid ständig komischi fehler wege concurrency macht als e zwar schnellere aber nid so reliable build. Zumindest isch das mol dr grund gsi, warum mer uf em CI bewusst dr daemon deaktiviert hän. Isch so viel stabiler gsi.

davidlahner (approved):
It's worth a try.

.circleci/config.yml
danielsuter: I think Gradle has improved a lot over the recent years. It's even recommended to use the daemon on CI. We're using it in our customer project without issues. If issues should arise, we can always switch back.

PR #1043, opened 2021-06-25
Improve Version Handling in Build & Backend

bruderol:
Improves the version handling in several aspects:
1. Works now when working with Spring Boot deployments - which seemed not to work before (could not load it from servlet context somehow)
2. refactored the version stuff into own gradle file to have all version stuff in one place
3. use resource processing of gradle to create version.properties file more elegantly
4. use same mechanism to also display version information as part of spring application banner in server log
5. use standard gradle version property in gradle files where appropriate (instead of deprecated version property or long scenarioo specific version property)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,39 +1,7 @@
-/**
- * Used Version of Scenarioo Java Library for reading scenarioo docu data
- */
-ext.scenariooApiVersion = '2.1.1'
+apply from: "version.gradle"
 
-/**
- * Internal scenarioo format version of produced aggregation data (internal server format version).
- *
- * On import this version is stored for each build, if the format of a build does not correspond to this format, the
- * build is automatically reimported again (all aggregation data recalculated in new format).
- *
- * First part of the version corresponds to library version that is supported, second part of the version depends on
- * internal aggregation format. the second part should be increased whenever something important is changed in the
- * internal format or the way that the aggregator is caluclating internal data on builds.
- */
-ext.scenariooAggregatedDataFormatVersion = '2.1.0'
-
-/**
- * Name of the release branch of this scenarioo version, which is used for links to the versioned documentation.
- */
-ext.documentationVersion = '5.0'
-
-/*
- * Gets the version name from the latest Git tag
- * From: http://ryanharter.com/blog/2013/07/30/automatic-versioning-with-git-and-gradle/
- */
-def getVersionName = { ->
-	def stdout = new ByteArrayOutputStream()
-	exec {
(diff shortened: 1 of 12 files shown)


build.gradle
```diff
@@ -1,39 +1,7 @@
...
-	exec {
-		commandLine 'git', 'describe', '--tags'
-		standardOutput = stdout
-	}
-	return stdout.toString().trim()
-}
+apply from: "version.gradle"
 
-ext.versionWithGitCommit = getVersionName();
+apply plugin: 'idea'
```
danielsuter, line 3: why do we need this?

scenarioo-server/build.gradle
```diff
@@ -96,55 +96,47 @@ dependencies {
...
-        def documentationVersion = project.documentationVersion
-
-        File versionFile = new File(sourceSets.main.output.classesDirs.getSingleFile().absolutePath + '/version.properties');
-        versionFile.write('version=' + versionWithGitCommit + '\n' +
-                'build-date=' + new Date() + '\n' +
-                'apiVersion=' + apiVersion + '\n' +
-                'aggregatedDataFormatVersion=' + aggregatedDataFormatVersion + '\n' +
-                'documentationVersion=' + documentationVersion + '\n')
+/**
+ * Replace properties in important resource files
```
danielsuter, line 100: I think java doc style should only be used on methods. I would use a /* single comment */ style here.

scenarioo-server/build.gradle
```diff
@@ -160,7 +152,7 @@ publishing {
 
             groupId 'org.scenarioo'
             artifactId 'scenarioo-viewer'
-            version versionWithGitCommit
+            version scenariooViewerVersion
```
danielsuter, line 155: this is the same as archiveVersion, correct?

scenarioo-server/src/test/java/org/scenarioo/dao/version/ApplicationVersionHolderTest.java
```diff
@@ -0,0 +1,17 @@
+package org.scenarioo.dao.version;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.*;
+
+public class ApplicationVersionHolderTest {
+
+	@Test
+	public void initializeFromClassContext_loadsVersionProperlyFromVersionPropertiesFile() {
```
danielsuter, line 10: JUnit 5 does not require the public keyword

danielsuter (changes requested):

